Issue #16, opened 2014-03-06
Example app for Camera facade


brousch:
We need a camera example app
##

kamathln:
@trivedigaurav Are you in India? I get this 
![screenshot](https://cloud.githubusercontent.com/assets/241043/17829754/846e952a-66d6-11e6-9bbf-3d242a1d3710.png)

jchome:
Hello, Now, source files are available again on this URL:
http://julien.coron.free.fr/raw/takepicture-plyer/

KeyWeeUsr:
Closed by #41

PR #41, opened 2014-03-13
Camera example

trivedigaurav:
This is the basic example for the camera facade. Saves the image with the name given by the user.

### DIFF
--- a/examples/camera/basic/README
+++ b/examples/camera/basic/README
@@ -1,0 +1,19 @@
+Basic camera example.
+Default picture is saved as /sdcard/org.test.cameraexample/enter_file_name_here.jpg
+
+
+Python-for-android compilation:
+
+	./distribute.sh -m 'kivy plyer'
+	cd dist/default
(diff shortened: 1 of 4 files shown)


redfast00:
@trivedigaurav I have the same problem as @amalgamatedclyde (pressing checkmark does nothing), but when I return to the application (by pressing X), I get a 'could not save picture' popup. Cyanogenmod 13.0 (Android 6) on OnePlus Two.

kiok46:
@redfast00  I will look into this in a couple of days (Have exams now).

redfast00:
@kiok46: Would you mind if I opened an issue?

dessant:
@redfast00, if you're convinced the bug is not in this pull request, but in plyer, then an issue whould be ok. Otherwise sorting it out on this thread is recommended.

kiok46:
@redfast00  Also, there is already an issue opened by @brousch  demanding a working example for camera façade. #16

akshayaurora:
@thegrymek thanks

PR #43, opened 2014-03-16
Added orientation facade, example and android implementation

inclement:
I'm not sure exactly how to construct this one, but I went for a (hopefully) fairly generic api based loosely on some of android's functions.

I'm not sure if this depends too much on exactly how android exposes these options (so it might not fit ios well), but on the other hand I like the idea of supporting as much of android's own api as possible. Anyway...comments appreciated!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,4 +27,5 @@ Notifications                      X             X                 X       X   X
 Text to speech                     X             X                 X       X   X
 Email (open mail client)                         X
 Vibrator                                         X
+Orientation                                      X
 ================================== ============= ============= === ======= === =====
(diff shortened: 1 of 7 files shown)


thegrymek:
Hi, 

I tested that on S Galaxy I with android 2.3(api 9) and another one with v. 4.4.5. On both phones it works perfectly. 

Do not for what there is a label: `Red functions need api level 18 if running on android!`. What functions?

best regards

dessant:
Updated in https://github.com/kivy/plyer/pull/135. Thanks @inclement. :+1:

PR #44, opened 2014-03-16
Osx win linux camera facade

trivedigaurav:
Added camera implementation using Kivy. It builds an app to capture and save an image from the camera. Similar to what we had for Android. As of now, Kivy's image.save() needs a patch wrt https://github.com/kivy/kivy/issues/1974, but fixing that makes this facade work fine.


akshayaurora:
Closing this as player is supposed to not be tied to any one UI library.

PR #64, opened 2014-06-14
Osx camera

trivedigaurav:
OSX Camera using the ImageSnap utility (http://iharder.sourceforge.net/current/macosx/imagesnap/).


akshayaurora:
Closing this as this uses a binary, plyer is supposed to be simple enough that it can just be included in iOS and android like mobile.

Including  a binary like this would be against that philosophy.

PR #68, opened 2014-06-28
Linux accelerometer facade

trivedigaurav:
Cleaned up version of @Davideddu's pull request: https://github.com/kivy/plyer/pull/51


Depau:
This should be re-written using the SensorProxy API:

https://developer.gnome.org/iio-sensor-proxy/stable/gdbus-net.hadess.SensorProxy.html

KeyWeeUsr:
@Depau if the current implementation returns the correct results and is only marked as deprecated (or only as old or something), feel free to add a new implementation, but leave the current one in place as a fallback or add the reasoning for using dbus, please.

I got 503 from that website. :/

PR #96, opened 2014-10-01
Camera implementation for Linux

Depaulicious:
Based on an ad-hoc camera app I made: https://github.com/Davideddu/kivy-camera
Kivy Camera needs to be installed system-wide, as the plyer module will look for its wrapper in the path.

It just runs the camera app using subprocess, then calls the callback.

I decided to make a new camera app because the existing one didn't have a feature that plyer requires, which is taking a picture, saving it to a specific path and exiting. The app should also work as a normal camera app (I didn't test it).

This should work on Windows and Mac OS X too, however kivy-camera's makefile would need some fixes in order to put the files in the right places in those platforms. Please pull request it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ Support
 Platform                           Android < 4.0 Android > 4.0 iOS Windows OSX Linux
 ================================== ============= ============= === ======= === =====
 Accelerometer                      X             X             X           X   X
-Camera (taking picture)            X             X
+Camera (taking picture)            X             X                             X
 GPS                                X             X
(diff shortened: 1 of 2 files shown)


akshayaurora:
We already have other camera prs doing the same... adding implementation using kivy... issue is player is supposed to be independent of kivy. The is the reason why these prs have not been included.

Closing this as the camera the facade should just work without having to install external dependencies.

yaki29:
@akshayaurora Can I work on it using just opencv, it won't contain any dependency on kivy-camera.

PR #104, opened 2015-01-19
Add filechooser facade and support for Linux and Windows

Davideddu:
This adds another alternative implementation of native file chooser dialogs, and implementations for Windows (using win32all) and Linux (using either YAD, Zenity or KDialog, based on which one is installed and what desktop environment is running).


trivedigaurav:
That's a great idea. I will try it out. Thanks!

trivedigaurav:
This was closed because @Davideddu opened a newer PR - https://github.com/kivy/plyer/pull/105#issuecomment-70554110.

Issue #22, opened 2014-03-11
GPS Support for iOS


brousch:
Add support for the GPS facade to iOS
##

emilv:
Should this be closed?

KeyWeeUsr:
Closed by #116.

PR #116, opened 2015-03-03
iOS GPS support

laltin:
Added support for GPS in iOS. Fixes issue #22

Working example: (on iPhone 5 and iOS 8)
image


tito:
This is awesome! Except the naming (it should be IosGPS, but a class starting with a lower case is not pep8 compliant), but that can be fixed later.

Simple and working :)

Issue #54, opened 2014-05-15
AndroidGPS list GPS Provider but use hardcoded "gps"


murlock:
List all Location Providers but only use gps:

```
        providers = self._location_manager.getProviders(False).toArray()
        for provider in providers:
            self._location_manager.requestLocationUpdates(
                "gps",
                1000, # minTime, in milliseconds
                1, # minDistance, in meters
                self._location_listener,
```

it should be 
```
            self._location_manager.requestLocationUpdates(
                provider,
```

PR #117, opened 2015-03-05
Android gps.py: fixed location provider cycling

JimmyStavros:
Issue was first brought to light here: https://github.com/kivy/plyer/issues/54

GPS module was cycling through "gps" three times instead of all location services.


tito:
Thanks you!

PR #115, opened 2015-03-02
use environ to change LANG to 'C' while calling lshw

tshirtman:
fixes #114

PR #119, opened 2015-03-08
Use environ to change LANG to 'C' while calling shell processes

trivedigaurav:
Using the protocol described in https://github.com/kivy/plyer/pull/115 to call external shell processes. This is to prevent issues due to localized output.


tito:
Thanks :)

PR #120, opened 2015-03-09
add video recoding to Camera facade and camera.py

pspchucky:


tito:
Coool! I guess the README should be improved to include recording support or not. Well done :)

tito:
Also, maybe the quality parameter could be exposed as well

PR #122, opened 2015-03-14
pep8 - removed unused imports and variables

thegrymek:


plyer/platforms/win/libs/win_api_defs.py
```diff
@@ -9,11 +9,10 @@
            'DestroyWindow', 'LoadIconW')
 
 import ctypes
-from ctypes import (Structure, windll, sizeof, byref, POINTER, memset,
-                    WINFUNCTYPE)
+from ctypes import Structure, windll, sizeof, POINTER, WINFUNCTYPE
 from ctypes.wintypes import (DWORD, HICON, HWND, UINT, WCHAR, WORD, BYTE,
-    LPCWSTR, LPWSTR, INT, LPVOID, HINSTANCE, HMENU, LPARAM, WPARAM,
```
trivedigaurav: I think we can leave these ones here for this is a lib and maybe useful later...

thegrymek:
We can reimport it later. If it will be soon used for future implementation so should we create issue on github for that or todo inside code with some description.

Please correct me if im wrong.

tito:
Seems good to me, thanks :)

PR #125, opened 2015-03-18
android contact list example

thegrymek:
Hi,

I want to share android version of contact list. There's example, facade and android implementation.  
Could you review and check it on your phones?

Thanks!
best regards

### DIFF
--- a/plyer/facades.py
+++ b/plyer/facades.py
@@ -93,6 +93,39 @@ class Camera(object):
         raise NotImplementedError()
 
 
+class Contacts(object):
+    '''Contacts Facade
+
+    .. versionadded:: 1.2.4
+
+    '''
+
+    _contacts = []
+
+    def __len__(self):
+        return len(self._contacts)
+
+    def __getitem__(self, item):
+        return self._contacts[item]
+
+    def __iter__(self):
+        return self._contacts.__iter__()
+
+    def refresh(self):
+        """Refreshes local contact list"""
+        raise NotImplementedError()
+
+    def insert(self):
+        """Creates and inserts contact into system"""
+        raise NotImplementedError()
+
+    def get(self):
+        """Returns all contacts
+        :rtype: list
+        """
+        return self._contacts
+
+
 class Notification(object):
     '''Notification facade.
     '''
--- a/plyer/facades.py
+++ b/plyer/facades.py
@@ -67,7 +67,9 @@ class Camera(object):
         :type filename: str
         :type on_complete: callable
         '''
+
         self._take_picture(filename=filename, on_complete=on_complete)
+
     def take_video(self, filename, on_complete):
         '''Ask the OS to capture a video, and store it at filename.
 
@@ -82,13 +84,14 @@ class Camera(object):
         :type filename: str
         :type on_complete: callable
         '''
+
         self._take_video(filename=filename, on_complete=on_complete)
 
     # private
 
     def _take_picture(self, **kwargs):
         raise NotImplementedError()
-        
+
     def _take_video(self, **kwargs):
         raise NotImplementedError()
 
@@ -101,6 +104,10 @@ class Contacts(object):
     '''
 
     _contacts = []
+
+    def __init__(self):
+        """Using refresh that should full fil _contacts"""
+        self.refresh()
 
     def __len__(self):
         return len(self._contacts)
@@ -131,7 +138,7 @@ class Notification(object):
     '''
 
     def notify(self, title='', message='', app_name='', app_icon='',
-                timeout=10):
+               timeout=10):
         '''Send a notification.
 
         :param title: Title of the notification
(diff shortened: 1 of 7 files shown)


plyer/facades.py
```diff
@@ -82,23 +84,61 @@ def take_video(self, filename, on_complete):
...
 
     def _take_picture(self, **kwargs):
         raise NotImplementedError()
-        
+
     def _take_video(self, **kwargs):
         raise NotImplementedError()
 
 
+class Contacts(object):
```
tito: Documentation lack of knowing the exact format of the contact list, and usage. The attached example is too complex to understand, but demonstrate it.

Could you explain how a contact is structured? What field are always here, the one that are optional? Best would be to give an output example like:

    >>> from plyer import Contacts
    >>> Contacts.refresh()
    >>> print "I have {} contacts".format(len(Contacts))
    I have 3 contacts
    >>> print Contacts[0]
    { .... }

plyer/facades.py
```diff
@@ -82,23 +84,61 @@ def take_video(self, filename, on_complete):
...
+        return self._contacts[item]
+
+    def __iter__(self):
+        return self._contacts.__iter__()
+
+    def refresh(self):
+        """Refreshes local contact list"""
+        raise NotImplementedError()
+
+    def insert(self):
```
tito: Don't add it if it's not implemented at all.

plyer/facades.py
```diff
@@ -82,23 +84,61 @@ def take_video(self, filename, on_complete):
...
+        return len(self._contacts)
+
+    def __getitem__(self, item):
+        return self._contacts[item]
+
+    def __iter__(self):
+        return self._contacts.__iter__()
+
+    def refresh(self):
+        """Refreshes local contact list"""
```
tito: "Refresh the contact list"

plyer/facades.py
```diff
@@ -82,23 +84,61 @@ def take_video(self, filename, on_complete):
...
+
+    .. versionadded:: 1.2.4
+
+    '''
+
+    _contacts = []
+
+    def __init__(self):
+        """Using refresh that should full fil _contacts"""
+        self.refresh()
```
tito: You missed to call the object constructor as well: `super(Contacts, self).__init__()`

plyer/platforms/android/contacts.py
```diff
@@ -0,0 +1,67 @@
...
+
+            has_phone_number = contact_cr.getColumnIndex('has_phone_number')
+            has_phone_number = int(contact_cr.getString(has_phone_number))
+            contact['has_phone_number'] = has_phone_number
+
+            phone_numbers = []
+            if contact['has_phone_number'] > 0:
+                l = ArrayList()
+                l.add(contact['contact_id'])
+                query = String("contact_id=?")
```
tito: The query doesn't work with a simple python string?

plyer/platforms/android/contacts.py
```diff
@@ -0,0 +1,67 @@
...
+    .. versionadded:: 1.2.4
+
+    """
+
+    def refresh(self):
+        """Refresh local contact list."""
+        cr = activity.getContentResolver()
+        contact_cr = cr.query(JavaContacts.CONTENT_URI, None, None, None, None)
+        if contact_cr.getCount < 1:
+            return
```
tito: getCount is not a method? Also if there is no contact, you should reset the _contacts as well

thegrymek: should be, shame on me

plyer/platforms/android/contacts.py
thegrymek: I remember not, I will double check

plyer/facades.py
thegrymek: 1. These two links give you a glance to know how its structured:
http://developer.android.com/images/providers/contacts_structure.png

and scroll to the down to know how look tables:
http://xianminx.github.io/2012/06/24/android-contact-provider-in-deep/

Contacts and other data like emails or groups are in sqlite database. Everything we can get by doing a query.

2. print "I have {} contacts".format(len(Contacts))

I have added that feature

tito:
That's great! Now, you need to get a little step outside the implementation to look at the result, and ask yourself: is it good and simple enough? is the format will works into other platforms (iOS ?)

The latest is too close on the android implementation (with column, query etc.).
The API is 1:1 android query / field name, obvsiouly, this wont work on iOS the same, and will put all the work on translating / mapping to the iOS implementation

Now that i've seen the output format (is it the whole thing, or not? If you have any non-documented field, they MUST be documented. The user will not check the android nor ios documentation to see if a field exist or not.)

Let's take your example:
```
{
  'contact_id': 150,
  'display_name': 'Johny W'
  'has_phone_number': '2',
  'phone_numbers': [ '111-111-111', '111-111-112']
}
```

Field like "has_phone_number" is redundant. It has a meaning in a db (to prevent subquery), but here on a result, you can always have a "phone_numbers", and if you want to know how much, just use len(). 

Right now, best would be:

- List all the fields (they can have another name, it doesn't have to be the exact same as Android) you wish to expose in the Contacts API
- Document them
- Do an implementation for Android

Be platform-abstract :)

tito:
And forget about the query() API right now, focus on the what you want in field in the API, ensure the format / naming you choose would be ok for all the platform, then you focus on Android.

thegrymek:
As I looked up to other platforms I've noticed that only ID fields are required. Rest are optional, even display_name 

In doc, below of facade/Contacts class I listed only 3 fields because rest of all are not implemented. Should I add them anyway or add with implementation?

I didnt add any mapper to that commit, because I think it would be great for another topic for next pull request. Add my propose of "fields mapper" here or in the another branch?

Julian-O:
Would the reviewers care to give a status on this one? Can we close it or merge it?

PR #129, opened 2015-04-11
vibrator for android v < 4.0

thegrymek:
Feature `exists` is supported only for android SDK >= 11. Rest features work. 

Tested on Samsung Galaxy I, android 2.3.

What we should do? Does we need new Exception `NotSupported` or just return `None`
Currently it returns None if SDK < 11

best regards


thegrymek:
I added some docs there. Please review them and share a comments. Still learning how to do it well

examples/vibrator/main.py
```diff
@@ -1,11 +1,7 @@
-
+"""Plyer Vibrator Example."""
```
kived: Simple, one-line docstrings like this don't add any information and should be avoided. (We know this is a Plyer Vibrator Example, because it's examples/vibrator/main.py in Plyer. Similarly we know that VibrationApp is the Main Application, and that build() returns the root widget, etc., without needing docstrings.) Also, remember to follow our standards (look at the rest of the code). We use `''''''` for docstrings, not `""""""`.

plyer/platforms/android/vibrator.py
```diff
@@ -8,27 +11,43 @@
...
 
         if vibrator:
             vibrator.vibrate(pattern, repeat)
 
-    def _exists(self, **kwargs):
-        return vibrator.hasVibrator()
+    def _exists(self):
+        if SDK_INT >= 11:
+            return vibrator.hasVibrator()
+        return NotImplementedError()
```
thegrymek: Please review this line.

dessant:
:+1:

PR #133, opened 2015-04-23
fix #107 - Use Android_ID instead of IMEI

### LINKED ISSUE #107: AndroidUniqueID doesn't use Android ID
The implementation returns the IMEI and therefore only works on phones, not tablets. Android provides an Android ID via Secure Settings which works for all devices and does not require any extra permissions (the current implementation requires READ_PHONE_STATE).

Example:
````python
PythonActivity = autoclass('org.renpy.android.PythonActivity')
Secure = autoclass('android.provider.Settings$Secure')
android_id = Secure.getString(PythonActivity.mActivity.getContentResolver(),
                              Secure.ANDROID_ID)
````

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7940424-androiduniqueid-doesn-t-use-android-id?utm_campaign=plugin&utm_content=tracker%2F77151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F77151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

aron-bordin:
Using Android_ID. 
Do I need to add something else?

### DIFF
--- a/plyer/facades/uniqueid.py
+++ b/plyer/facades/uniqueid.py
@@ -3,15 +3,14 @@ class UniqueID(object):
 
     Returns the following depending on the platform:
 
-    * **Android**: IMEI
+    * **Android**: Android ID
     * **Mac OSX**: Serial number of the device
     * **Linux**: Serial number using lshw
     * **Windows**: MachineGUID from regkey
-
-    .. note::
-        On Android your app needs the READ_PHONE_STATE permission
-
+    
     .. versionadded:: 1.2.0
+
+    .. versionchanged:: 1.2.4 On Android returns Android ID instead of IMEI
     '''
 
     @property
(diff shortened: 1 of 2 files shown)


Kovak:
An important observation:

[14:20] <tito> use "from plyer.platforms.android import activity" instead of manually importing PythonActivity
[14:21] <tito> otherwise the code will be broken when used in service
[14:21] <tito> dessant: ie, use: https://github.com/kivy/plyer/blob/master/plyer/platforms/android/__init__.py

aron-bordin:
PR updated. Please, when possible, let me know if it's ok now.

dessant:
Looks great, but we missed to merge this before refactoring the facades. Could you please update it and add a versionchanged tag with a short mention of the change on android? It should be merged right after that.

aron-bordin:
PR updated, check if it's ok

dessant:
Thanks a lot. :+1:

PR #135, opened 2015-05-03
Inclement orientation

thegrymek:


plyer/facades/orientation.py
```diff
@@ -0,0 +1,43 @@
+class Orientation(object):
+    '''Orientation facade.
+
```
dessant: Let's add a versionadded tag.

dessant:
@inclement, @thegrymek thanks. :dancer:

PR #137, opened 2015-05-04
Added optional duration argument to _take_video

JimmyStavros:


tshirtman:
would be suitable to add the keyword argument to the interface too.

PR #144, opened 2015-05-11
Plyer audio for android with facade and example

thegrymek:
Hi,

Recently I finished with audio recording for android. It works for android Api>=9.

Please check it out!

best regards


plyer/facades/audio.py
```diff
@@ -0,0 +1,56 @@
+class Audio(object):
+    '''Audio Facade.
+
+    Used for recording audio.
+    Use method `start` to start record and `stop` for stop recording.
+    For hear, what you have just recorded use method `play`.
```
tshirtman: nit: "To hear " (no comma). But a comma would be nice after "recorded".

plyer/facades/audio.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Status will tell you about current job of the Audio.
```
tshirtman: don't put a capital at the biggining of the sentence if it's an api name, it can be confusing. Also, for more clarity, using "`" around api names will make them more distinguishable.

`status` will tell you about the…

plyer/facades/audio.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Default path for recording is set in platform implementation.
+
+    .. note::
+        You need android permissions: RECORD_AUDIO
```
tshirtman: "on android the `RECORD_AUDIO` permission is needed."

plyer/platforms/android/audio.py
```diff
@@ -0,0 +1,59 @@
...
+# Player Classes
+MediaPlayer = autoclass('android.media.MediaPlayer')
+
+
+class AndroidAudio(Audio):
+    '''Audio for android.
+
+    For recording audio we use MediaRecorder Android class.
+    For playing audio we use MediaPlayer Android class.
+
```
tshirtman: nit: empty line not needed

tshirtman:
This looks nice overall, i have minor issues with documentation, but the rest is ok.

About codecs, i agree it's a complex problem, i think the best we could do would be to offer another api to return available codecs so the programmer can look if their favorite one is in the list, a bit like the result of `mplayer --vc=help` or `ffmpeg -codecs`.

dessant:
:+1:

PR #149, opened 2015-05-20
android utils with facade and example

thegrymek:
Hi,

Tested only on android sdk = 21.
On the evening I ll test it on sdk==9 also.

It can:
- start/stop/scan wifi
- show/hide keyboard
- display information about available sensors
- give information whether is connection to internet and is enabled wifi.


thegrymek:
On sdk==9 works also.

best regards.

tshirtman:
I think there has been a misunderstanding here, you were told to create facades for all new api that you added, but you create a new set of apis, and put it into a new common facade, even if these are only losely related, if you put this logic further, any new feature could be added to the Utils facade, instead of as a separate facade, and it seems to me that you are doing a plyer project inside plyer, that is, a set of api that are abstracted for usage in other projects.

So this should be cut into a few different facades, Network (check/ask wifi connection), keyboard (though i'm not sure about this one, can you define how it would be used?), and sensors… looks redundant to the accelerometer/gyroscope/orientation facades, and if you have more to put here, you should create more facades for them, the way this was done seems to indicate the train of though was to wrap android APIs to use them, and not look at what other platforms others and how to provide a sensible subset of common interfaces, if you don't think this through, you can easily implement facades that makes no sense to implement on other platforms.

dessant:
Closing this pr, after refactoring, every facade should be submitted as a seaparate pr.

PR #150, opened 2015-05-20
linux audio

thegrymek:
Hi,

This pr provides audio for linux. It uses standard python library pyaudio

best regards,

### DIFF
--- a/plyer/platforms/linux/audio.py
+++ b/plyer/platforms/linux/audio.py
@@ -48,12 +48,9 @@ class AudioWorker(Process):
     def run(self):
         super(AudioWorker, self).run()
         signal = None
-        chances = 100
-        while not signal and chances > 0:
+        while not signal:
             signal = self.get_signal()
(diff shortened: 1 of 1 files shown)


plyer/platforms/linux/audio.py
```diff
@@ -0,0 +1,155 @@
...
+    'stop_playing': 2,
+    'start_recording': 3,
+    'stop_recording': 4,
+}
+
+
+class AudioWorker(Process):
+    '''Process for recording and playing audio files.
+
+    Rather wave files.
```
tshirtman: Rather? do you mean Raw? or Record? i have no idea what this means.

tshirtman:
The pyaudio dependency doesn't seem to be documented, can you also explain why this lib was checked in particular? did you consider other choices?

bhaveshAn:
Fixed this feature at #311 as not having dependencies, kindly view this.

PR #156, opened 2015-06-09
magnetic field example with facade android

thegrymek:
Hi,

Implementation for magnetic field for android with example and facade. I based on acceletometer implementation.

tests on api = 22 and api = 9

best regards,
Andrzej Grymkowski


thegrymek:
I close this one because there is implementation for it called Compass

PR #157, opened 2015-06-11
Plyer android orientation listener

thegrymek:
Orientation listener for android.

Updated example and facade.

Tested on sdk = 21 and sdk = 9

best regards,
Andrzej Grymkowski

### DIFF
--- a/examples/orientation/main.py
+++ b/examples/orientation/main.py
@@ -1,45 +1,104 @@
 
-from kivy.base import runTouchApp
-
+from kivy.app import App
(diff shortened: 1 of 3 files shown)


po9794:
I have implemented a new sensor called Rotation, that gets orientation (azimuth, pitch, roll), in a more precise and stable way. But i couldn't call it compass not orientation, so i called it Rotation, also based on the name of the high level sensor it uses in android api : TYPE_ROTATION_VECTOR.

malverick:
works fine in #291

PR #160, opened 2015-06-12
Plyer android pressure sensor

thegrymek:
Hi,

Example, facade and android implementation.

Both android and ios measures pressure in hPa and that unit we will see in example.

Best regards


thegrymek:
fixed type issues and tested on sdk=9 and sdk=21.

dessant:
Merged in https://github.com/kivy/plyer/pull/286.

PR #161, opened 2015-06-13
Plyer android gravity

thegrymek:
Example, facade and android implementation.

Tested on android sdk=9 and sdk=21


malverick:
works fine in pr #279

PR #162, opened 2015-06-20
Add rotation sensor

po9794:
Add a sensor to get Azimuth (usefull for a compass), Pitch and Roll.
Uses high-level sensor (so called sensor fusion), to transparently use the accelerometer, gyroscope and magnetometer if they are available. It needs to initially orient itself and then eliminate the drift that comes with the gyroscope over time.
The data are stable and we have a great response time.

### DIFF
--- a/examples/rotation/buildozer.spec
+++ b/examples/rotation/buildozer.spec
@@ -1,0 +1,173 @@
+[app]
+
+# (str) Title of your application
+title = Kivy Rotation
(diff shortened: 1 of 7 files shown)


tshirtman:
not tested yet, but looks good to me.

peterLaurence:
Tested on Android 5.1.1.
Works fine.

PR #170, opened 2015-08-03
Gallery facades for android and iOS along with examples.

akshayaurora:


Mr-Victor:
I have added the changed files to plyer, I ran the example and it opened. I selected "Choose from gallery" and it opened the native gallery on Android. After image selection it closed so I added the "on_pause" function and it returned to the app after image selection. I ran with logcat but cant seem to find the problem. Using prints in the "plyer/platforms/android/gallery.py" I realized "data = intent.getClipData()" is None after selection. I cant seem to resolve this. Any help will be appreciated.

Issue #175, opened 2015-10-31
Something wrong with encoding in AndroidNotification


Hukuta:
```
def notify(title, message):
    return AndroidNotification().notify(title=title, message=message)

notify(u'Vk Service', u'Нет новых сообщений')
```

and i see this notification:

![s51101-014425](https://cloud.githubusercontent.com/assets/3100053/10866530/fa84a610-803a-11e5-85cf-69bc10c76b8a.jpg)

but next works without such problems:

```
class MyApp(App):
    def build(self):
        return Label(text=u'Привет')
```

![2015-10-31 23 54 50](https://cloud.githubusercontent.com/assets/3100053/10866542/a5a3b892-803b-11e5-9b0a-1c2ca5240970.jpg)

What can I do to localize notifications? 
##

Kovak:
Make sure the encoding on your supplied strings is utf-8 as that is what plyer is expecting.

Hukuta:
I'm using unicode strings ( u'...' )
Look at method _notify in [AndroidNotification](https://github.com/kivy/plyer/blob/master/plyer/platforms/android/notification.py) class, 
I see ".encode('utf-8')" there: 

    def _notify(self, **kwargs):
        icon = getattr(Drawable, kwargs.get('icon_android', 'icon'))
        noti = NotificationBuilder(activity)
        noti.setContentTitle(AndroidString(
            kwargs.get('title').encode('utf-8')))
        noti.setContentText(AndroidString(
            kwargs.get('message').encode('utf-8')))
        noti.setSmallIcon(icon)
        noti.setAutoCancel(True)

        if SDK_INT >= 16:
            noti = noti.build()
        else:
            noti = noti.getNotification()

        self._get_notification_service().notify(0, noti)

double "encode" leads to exception:

![selection_140](https://cloud.githubusercontent.com/assets/3100053/10867898/180dd916-8083-11e5-9586-57bf8eb76c22.png)

Hukuta:
I made [demo app](https://github.com/Hukuta/kivy-plyer-issue-175) to demonstrate problem, may be this will help to solve it.

IvanTheFirst:
I had problem with sms example, maybe this is the same? Try this:
unicode(str('я').strip(codecs.BOM_UTF8), 'utf-8')

this will clean utf8 header from the Python string

KeyWeeUsr:
[...]

PR #176, opened 2015-11-12
#175 fixing problems with encoding

Hukuta:
#175 fixing problems with encoding


Hukuta:
This solves https://github.com/kivy/plyer/issues/175

doratoa:
I believe it was already resolved by now? Should we close it?

PR #183, opened 2015-12-31
Adding battery example, notification ticker and gps example update

kiok46:
Here is the screen shot of the battery example on android device.

screenshot_2015-12-31-20-10-27

And a screenshot showing the notification ticker (at the top).

scre


plyer/facades/notification.py
```diff
@@ -2,8 +2,8 @@ class Notification(object):
     '''Notification facade.
     '''
 
-    def notify(self, title='', message='', app_name='', app_icon='',
-                timeout=10):
+    def notify(self, title='', message='', ticker='', app_name='', app_icon='',
+               timeout=10):
```
tshirtman: Adding a new keyword argument should happen at the end of the argument list, or people using them with direct values ("app name" instead of app_name="app name") will get a breakage.

plyer/facades/notification.py
```diff
@@ -17,8 +17,8 @@ def notify(self, title='', message='', app_name='', app_icon='',
         :type app_icon: str
         :type timeout: int
         '''
-        self._notify(title=title, message=message, app_icon=app_icon,
-                     app_name=app_name, timeout=timeout)
+        self._notify(title=title, message=message, ticker=ticker,
+                     app_icon=app_icon, app_name=app_name, timeout=timeout)
```
tshirtman: same.

kiok46:
@tshirtman  updated!, please check if it's ok. :smile:

tshirtman:
Looks good to me :) :+1:

PR #184, opened 2016-01-24
bluetooth for android and linux

kiok46:
Adding start, stop and turn on visiblity of bluetooth  for android and linux.
Moto e
screenshot_2016-01-25-01-54-15
screenshot_2016-01-25-01-54-06
Ubuntu 14.04
blue

I plan to implement discovering devices and pairing in future.

### DIFF
--- a/plyer/platforms/linux/bluetooth.py
+++ b/plyer/platforms/linux/bluetooth.py
@@ -1,0 +1,22 @@
+from subprocess import Popen, PIPE
+from plyer.facades import Bluetooth
+from plyer.utils import whereis_exe
+import subprocess
+from os import environ
+
+
+class LinuxBluetooth(Bluetooth):
+
+    def _start(self):
+        subprocess.Popen(["service", "bluetooth", "start"])
+
+    def _stop(self):
+        subprocess.Popen(["service", "bluetooth", "stop"])
+
+
+def instance():
+    import sys
+    if whereis_exe('bluetoothctl'):
+        return LinuxBluetooth()
+    sys.stderr.write("bluetoothctl not found.")
+    return Bluetooth()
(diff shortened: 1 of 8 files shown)


plyer/platforms/linux/bluetooth.py
```diff
@@ -0,0 +1,22 @@
...
+from plyer.facades import Bluetooth
+from plyer.utils import whereis_exe
+import subprocess
+from os import environ
+
+
+class LinuxBluetooth(Bluetooth):
+
+    def _start(self):
+        subprocess.Popen(["service", "bluetooth", "start"])
```
kived: An app should not be starting/stopping a system service. Doing this also requires root access. The specific command here is based on the Upstart init system, which is used in older Debian/Ubuntu systems and has since been replaced by systemd. While the Upstart commands still work for backwards compatibility, I wouldn't count on that always being the case. (In fact, systemd is the only reason this command works now. Upstart's `service' would not have asked for a password, and MUST be run via sudo/gksudo instead.)

This also doesn't actually turn Bluetooth on/off, it just pairs/unpairs with preconfigured devices. `rfkill block`/`unblock` should be used to actually turn BT on/off. It should not require root access if the user has appropriate permissions.

tito:
I think the API Bluetooth.on_stop is not good. What you want is unload or cleanup method instead, to unregister anything that registered during enable.

After thinking a little bit more, it make more sense to have disable() unregister here, since enable() register.

kiok46:
@tito   I think I should register and unregister in onResume and onPause method.  
onResume() is always called on the activity being displayed.
also see here first answer here. : http://stackoverflow.com/questions/7439041/how-to-unregister-broadcastreceiver

tito:
Yeah, but here i think that's implementation issue on your provider. Maybe pause/resume in the generic API then? or register/unregister. But at least not on_stop().

Also be carefull about multiple call of enable() / disable() / enable() -> android register will be called twice.

PR #188, opened 2016-03-08
Android toast

kiok46:
Android: Moto e Screenshots

screenshot_2016-03-09-03-48-27
screenshot_2016-03-09-03-48-17


dessant:
Toast was a controversial topic and it was rejected in the past, because there is no clear equivalent on other platforms, and that naturally makes the feature unsuitable for plyer.

Issue #178, opened 2015-12-03
Wrong SMS example, app crashes when your try to use unicode symbols


IvanTheFirst:
Hello, 
I think there is some error in sms example ( https://github.com/kivy/plyer/tree/master/examples/sms ).
Your should use
`smstext = unicode(self.sms_message.strip(codecs.BOM_UTF8), 'utf-8')
sms.send(recipient=self.sms_recipient, message=smstext)`

instead of 
`
    sms.send(recipient=self.sms_recipient, message=self.sms_message)
`
For some languages your need to use utf16.

https://docs.python.org/2/howto/unicode.html 
> Unicode character U+FEFF is used as a byte-order mark (BOM), and is often written as the first character of a file in order to assist with autodetection of the file’s byte ordering. Some encodings, such as UTF-16, expect a BOM to be present at the start of a file; when such an encoding is used, the BOM will be automatically written as the first character and will be silently dropped when the file is read. There are variants of these encodings, such as ‘utf-16-le’ and ‘utf-16-be’ for little-endian and big-endian encodings, that specify one particular byte ordering and don’t skip the BOM.

Julian-O:
Note: Python 2.7 support is no longer required. Call to `unicode()` may need to be reconsidered.

PR #189, opened 2016-03-09
pyler/main: Unicode symbols fix

mr-karan:
Changed the message body to allow for Unicode symbols in SMS message

Fixes: https://github.com/kivy/plyer/issues/178

### DIFF
--- a/examples/sms/main.py
+++ b/examples/sms/main.py
@@ -42,8 +42,9 @@ class IntentButton(Button):
     sms_message = StringProperty()
 
     def send_sms(self, *args):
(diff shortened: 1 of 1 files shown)


dessant:
`unicode` will not work under python 3.

kived:
If there is an issue, then the issue is with the implementation (for one or more platforms) and not with the example. Either way, this is the wrong fix. As @dessant said, `unicode` does not work in py3. The SMS facades should accept either utf-8 encoded bytes or a unicode/py3 str object and work no matter what.

PR #190, opened 2016-03-11
Corrected a bug in the file chooser when running in Python 3.

eje211:
When running the file chooser in Python 3, I got a TypeError, "Cannot write str to text stream." This change solved the issue for me.

### DIFF
--- a/plyer/platforms/linux/filechooser.py
+++ b/plyer/platforms/linux/filechooser.py
@@ -5,7 +5,7 @@ Linux file chooser
 
 from plyer.facades import FileChooser
 from distutils.spawn import find_executable as which
-import os
+import os, sys
 import subprocess as sp
 import time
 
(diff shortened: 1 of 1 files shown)


plyer/platforms/linux/filechooser.py
```diff
@@ -63,7 +63,9 @@ def _split_output(self, out):
         '''This methods receives the output of the back-end and turns
         it into a list of paths.
         '''
-        return out.split(self.separator)
+        if sys.version_info[0] == 2: return out.split(self.separator)
+        else: return out.decode(sys.getfilesystemencoding()).split(self.separator)
```
tshirtman: please avoid inlining if/else blocks.

seems good overwise.

dessant:
There's also a compat module, you could import `PY2` from there: https://github.com/kivy/plyer/blob/master/plyer/compat.py#L16.

KeyWeeUsr:
Closed by https://github.com/kivy/plyer/commit/6e1f86c75fe9391dad50c752c1d176584e4c86b1.

PR #192, opened 2016-03-20
Rewrite notification on Mac using PyOBJus

andong777:
I rewrite the notification feature on mac using PyOBJus. 
I have tested it on my MacBook, and it works fine.
The original code is written two years ago, and it seems it doesn't work. Besides, it uses  PyObjC instead of PyOBJus.


akshayaurora:
@andong777 nice work!

PR #194, opened 2016-03-22
In-app browser: example, facade, and iOS naive implementation

andong777:
I made an attempt to implement a new feature, in-app browser. The in-app browser, as I know, is a view(, view controller, activity) to let user browser something within the app. It currently owns one simple method `open_url`, and more methods, properties can be added if our requirement grows.
The example app I designed contains only a text input for inputting url and a button. I don't know whether there is the concept 'in-app browser' on desktop, and I only implement it on iOS as a try. 
The implementation on iOS is naive, and actually it is not internal.

`UIApplication.sharedApplication().openURL_(nsurl)`

Anyway, I have tested it on iOS 9, iPad mini 2 and it works.

I have come up with how to implement it in a properer way, but a few things need to be done. Maybe I will do it in the next PR:
I want to import a third-party class, which can be easily used by PyOBJus. But I don't know if it is allowed to add new dependency for this feature.
As Apple restricts in-app HTTP request since iOS 9 [(ref),](http://stackoverflow.com/questions/31254725/transport-security-has-blocked-a-cleartext-http) info.plist needs a minor modification to allow a real in-app browser. I'm not sure if it is user-friendly.

### DIFF
--- a/examples/inappbrowser/buildozer.spec
+++ b/examples/inappbrowser/buildozer.spec
@@ -1,0 +1,173 @@
+[app]
+
+# (str) Title of your application
+title = Plyer In-app browser Example
+
+# (str) Package name
+package.name = inappbrowser
+
+# (str) Package domain (needed for android/ios packaging)
+package.domain = org.test
+
+# (str) Source code where the main.py live
+source.dir = .
+
+# (list) Source files to include (let empty to include all the files)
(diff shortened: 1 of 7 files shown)


akshayaurora:
Currently this just implements open_url opening a external browser through it. Neither is it an actual in app browser, nor is it giving user control of the browser. Please open after fixing these issues.

PR #196, opened 2016-04-14
Py3 compatibility, open browser and edit sms

emillynge:
Hi, I have several features in this PR

* I have fixed a bug that caused py3 to crash on android
* pyjnius calls to `PythonActivity` in have been redirected use `activity` from `android.__init__` where found. the autoclass of `PythonActivity` tries to use `kivy.setupconfig.USE_SDL2` to determine whether `org.renpy` or `org.kivy` should be used, as `org.renpy` is deprecated and can cause crashes.
 * Also note that imports from `plyer.platform.android.__init__` have been refactored to `from . import ...` as I find it much cleaner and futureproof. This can easily be changed back ;)
* A `Browser` has been implemented for linux and android. It just opens the platform native browser - useful for clicking links.
* An `edit` method has been added to `Sms` class for android. It opens the native SMS application and starts a message that can be populated with:
 * recipient
 * message
 * recipient _and_ message

I did not mean to bunch everything into a single PR. But I was just using it for my own application, and suddenly I had implemented several things I might as well share. Unless you guys have a problem with one or more of the features I'm not inclined to bother trying to separate them into individual PRs.

Regards 
Emil Lynge

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,8 +27,9 @@ GPS                                X       X
 Notifications                      X           X       X    X
 Text to speech                     X       X   X       X    X
 Email (open mail client)           X       X   X       X    X
+Browser (open browser client)      X                        X
 Vibrator                           X       X
-Sms (send messages)                X
+Sms (send messages or open client) X
 Compass                            X       X
 Unique ID                          X       X   X       X    X
 Gyroscope                          X       X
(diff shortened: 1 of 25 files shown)


doratoa:
Opening browser is already works with python's built-in _webbrowser_ module on all platforms (except it's broken on android_new as I heard) If other commits stull apply I supposed they should be moved into separate PR(s) ?

PR #197, opened 2016-04-17
Implementing brightness for linux

kiok46:
This PR implements Brightness for linux.
**User can do the following:**
- Change the brightness of the screen (0-100) **[set_brightness]**
- Increase brightness by some percentage. **[inc_brightness]**
- Decrease brightness by some percentage. **[dec_brightness]**
- Get the brightness value and use it as per need. **[get_brightness]** 

I have made use of [xbacklight](http://www.x.org/archive/X11R7.5/doc/man/man1/xbacklight.1.html).
**I have tested it and it works smoothly on ubuntu 14.04.**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,6 +21,7 @@ Supported APIs
 Platform                           Android iOS Windows OS X Linux
 ================================== ======= === ======= ==== =====
 Accelerometer                      X       X           X    X
+Brightness                                                  X
 Call                               X       X
 Camera (taking picture)            X       X
 GPS                                X       X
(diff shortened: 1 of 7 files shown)


kiok46:
could be merged.

akshayaurora:
If you are going to implement a feature it needs to be implemented in all platforms. except in cases where it might not be feasible.

akshayaurora:
Try to use something like ACPI that would be universal rather than a binary.
https://wiki.archlinux.org/index.php/backlight

KeyWeeUsr:
Unfortunatelly the same thing with `xbrightness` has been implemented and merged in #346. I'll take a look at the `backlight` kernel sysclass and remove the binaries approach later.

PR #201, opened 2016-04-30
Sysinfo

kiok46:
Sysinfo:
Currently displays details like: platform, processor, version, os name and more to be added in future.
Right now, I have only used "platform" module of python.


plyer/platforms/win/libs/screenmetric.py
```diff
@@ -0,0 +1,10 @@
+__all__ = ('get_SystemMetrics')
```
kiok46: is it ok to create this file? or should I put this function in `win_api_defs.py` and then in `win/sysinfo.py` use `plyer.platforms.win.libs import win_api_defs` instead of `plyer.platforms.win.libs import screenmetric`.

plyer/platforms/win/sysinfo.py
```diff
@@ -0,0 +1,64 @@
...
+class WindowsSysinfo(Sysinfo):
+
+    values = {}
+
+    def __init__(self, **kwargs):
+        self.SysInfo()
+
+    def SysInfo(self):
+
+        cache = os.popen2("SYSTEMINFO")
```
kiok46: Should I create a thread here? because application takes noticable amout of time to load because of this.

plyer/platforms/win/libs/screenmetric.py
kived: If all this does is call the function from the other module, then no, there isn't any reason to have this module.

plyer/platforms/linux/sysinfo.py
```diff
@@ -0,0 +1,66 @@
...
+from subprocess import Popen, PIPE
+from plyer.facades import Sysinfo
+
+
+class LinuxSysinfo(Sysinfo):
+
+    def _model_info(self, **kwargs):
+        password = kwargs.get('password')
+        command = 'dmidecode -s baseboard-product-name'.split()
+        p = Popen(['sudo', '-S'] + command, stdin=PIPE, stderr=PIPE,
```
kived: You shouldn't be calling `sudo` here. The user's password won't be available to the app dev, and shouldn't be required to be entered into the program. You can get this information without root if you don't use dmidecode (and dmidecode may not be installed anyway). Also, `baseboard-product-name` isn't what you want here, but probably `system-product-name`.

```
ryan@ryan-mk3:~$ sudo dmidecode -s baseboard-product-name
1854
ryan@ryan-mk3:~$ sudo dmidecode -s system-product-name
HP 2000 Notebook PC
ryan@ryan-mk3:~$ cat /sys/devices/virtual/dmi/id/product_name 
HP 2000 Notebook PC
ryan@ryan-mk3:~$ sudo dmidecode -s system-manufacturer
Hewlett-Packard
ryan@ryan-mk3:~$ cat /sys/devices/virtual/dmi/id/sys_vendor 
Hewlett-Packard
```

plyer/platforms/android/sysinfo.py
```diff
@@ -0,0 +1,106 @@
...
+        elif version >= 4.4.2 and version <= 4.4.4:
+            return ("Android", version, "KITKAT_MR2")
+        elif version >= 5.0 and version < 5.1:
+            return ("Android", version, "LOLLIPOP")
+        elif version >= 5.1 and version <= 5.1.1:
+            return ("Android", version, "LOLLIPOP_MR1")
+        elif version == 6.0:
+            return ("Android", version, "MARSHMALLOW")
+        else:
+            return ("Android", 0.0, "Unknown version")
```
kived: You should be able to do this by examining Build.VERSION_CODES. Something like:

```
sdkint = BuildVersion.SDK_INT
for name, value in dir(BuildVersionCodes):
    if sdkint == value:
        return name
return 'UNKNOWN'
```

This is just an untested example, but doing it this way means it doesn't need to be updated for new versions - it will just work. Might need some modification, like checking that `value` is actually an `int`.

plyer/platforms/win/sysinfo.py
kived: No, I would suggest deferring this call. Make it check if `self.values` is set, and return immediately if it is, then you can call this in every other method before accessing `self.values`. Also, please remember to follow naming conventions, and also this is a private method, so calling it something like `_get_sysinfo` or `_ensure_sysinfo` would be better.

kiok46:
Have to do testing for iOS and android, and have to complete processor method in iOS. 
otherwise it's complete. 💃

plyer/facades/sysinfo.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def kernel_version(self):
+        # returns the kernel name.
+        return self._kernel_version()
+
+    def storage_info(self):
+        # returns the storage capacity o fthe system.
+        return self._storage_info()
+
+    def screen_dimension(self):
```
dessant: Please rename it to screen_resolution, dimension is rather uncommon in this context.

kiok46: will do!

kiok46: ok, done (y)

plyer/platforms/ios/sysinfo.py
```diff
@@ -0,0 +1,102 @@
...
+        '''
+        return (self.device.systemVersion, " ", " ")
+
+    def _architecture_info(self):
+        '''
+        Returns the architecture in a tuple for example:
+        '''
+        sizeClass = autoclass('SizeClass')
+        instance = sizeClass.alloc().init()
+        bit = instance.get_bit_size()
```
kiok46: I need some alternative way to get this info.
`make_dylib` only supports OSX and not iOS, so I have to remove that `sysinfo.m` file.

Previous conversation with @kived  https://botbot.me/freenode/kivy/2016-06-10/?msg=67676984&page=3

plyer/platforms/ios/sysinfo.py
```diff
@@ -0,0 +1,102 @@
...
+        '''
+        Returns the device name for example:
+        '''
+        return self.device.name
+
+    def _manufacturer_name(self):
+        '''
+        Returns the manufacturer's name for example:
+        '''
+        return "Apple Inc."
```
kiok46: Can't find any `official` place from where I could get this name. so I had to write this down manually.

plyer/platforms/ios/sysinfo.py
```diff
@@ -0,0 +1,102 @@
...
+        '''
+        processinfo = NSProcessInfo()
+        return processinfo.operatingSystemVersionString
+
+    def _storage_info(self):
+        '''
+        Returns the amount of storage (RAM) in GB. for example:
+        '''
+        Storage = autoclass('StorageClass')
+        instance = Storage.alloc().init()
```
kiok46: Same as under `_architecture_info` method.

doratoa:
@kiok46 , I've added some fixes for android and some notes also. Feel free to comment or further improve #241

kiok46:
@doratoa Thanks for the changes, sure I will try to improve this

doratoa:
@kiok46 if you will be working on this, please post in any of PR threads, I want to meddle with it a little this week too. Nice PR, want it to be merged soon ^_^ I'll try make some more tests too

PR #205, opened 2016-05-16
Added camera example

redfast00:
Added a camera example based on the code in https://github.com/kivy/plyer/pull/41 . Instead of saving to the current working directory, the app now saves pictures to `app.user_data_dir`. This ensures that we can write to the destination.

### DIFF
--- a/examples/camera/buildozer.spec
+++ b/examples/camera/buildozer.spec
@@ -1,0 +1,230 @@
+[app]
+
+# (str) Title of your application
+title = Kivy Camera
+
+# (str) Package name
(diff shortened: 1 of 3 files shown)


WTRipper:
Ehm sorry but why do we have to write at app.user_data_dir ? I mean theoretically it should be possible to write on external storage as well if permissions were given right? Is it a bug? Is there something new or a fix?

redfast00:
CI fails because there is a line that is slightly longer than 80 characters, all other code is PEP 8.

PR #208, opened 2016-06-09
initial update

kiok46:
Implementation of Tags, and P2P (NFC) for android.


kiok46:
I am removing the card emulation part as that will take a lot more time then implementing other modes.
I will open a separate PR for card emulation after this is merged.

PR #211, opened 2016-06-25
add a generic kivy-based camera implementation for linux

antocuni:
This PR depends on this kivy PR/feature request: https://github.com/kivy/kivy/pull/4430
Without the Kivy patch you can invoke the camera only once, because it is never freed and cannot be re-opened a second time


akshayaurora:
Player has been deliberately kept independent of kivy so it can be used with any framework and not just kivy. This would alter that.

Issue #221, opened 2016-07-21
battery.status isCharging always shows false


Warlord77:
>>> battery.status
{'percentage': 88.0, 'isCharging': False} [when disconnected]
>>> battery.status
{'percentage': 88.0, 'isCharging': False} [when charging]
OS : Ubuntu 16.04 LTS x86_64

kiok46:
It's probably an issue with upower or ubuntu.

In the top it's clear that it's not charging but still down in the terminal it says `power supply: yes` and also in the `Power Statistics: Device Information`

Maybe it's because I am using it in virtual box, but the problem is not in the code. (although it should return True as it's always yes.)

<img width="821" alt="screen shot 2016-07-22 at 12 01 09 am" src="https://cloud.githubusercontent.com/assets/7335120/17034291/ab21a688-4f9f-11e6-8b19-999f9003ccb2.png">

PR #223, opened 2016-07-22
Fixed issue #221

Warlord77:
UPower daemon  does not have a option called power supply instead in "state"  it shows the state weather its charging or not.


akshayaurora:
@Warlord77 thanks

PR #227, opened 2016-08-02
Updated renpy reference (deprecated) to kivy

erm3nda:


inclement:
I think this is actually already supposed to work, that's what the JAVA_NAMESPACE part is for, though I guess it doesn't work if you don't have the android module. This solution therefore isn't quite right, as the renpy class is the appropriate fallback if the new one isn't available. We could try to have a better check for it though.

Thanks for the contribution though, and if I'm wrong about that feel free to reopen the PR.

erm3nda:
mmm, i've edited it because didn't work on my Samsung Y (armeabi v6). Maybe
my phone is so sold, but changing that line made it work finally. Android
2.3.6 minapi 9.  I still have to try out with armeabi-v7a (if that makes
any difference because the phone internal libs).

At least i can patch build folder before create dist files. Thank you.

~~ Self signed ~~


2016-08-02 23:37 GMT+02:00 Alexander Taylor <notifications@github.com>:

PR #232, opened 2016-09-19
Fix TypeError if `LANG` is not set in on osx

ForeverWintr:
In macosx.uniqueid, `$LANG` is retrieved with `old_lang = environ.get('LANG')`. This means that if `LANG` is not set, `old_lang` will be none. When plyer later tries to restore the original value of lang, `putenv` will complain that we're not supplying a string.  I've corrected this by popping the inserted `LANG` value if `old_lang` is none.


akshayaurora:
@ForeverWintr thanks!

PR #241, opened 2016-11-11
[WiP] Sysinfo (from #201)

doratoa:
* Based on #201 by @kiok46 
Adding some fixes for android implementation and example outputs.

- fix: android sysinfo wouldn't compile due to syntax error
- fix: **_screen_resolution()** was unfinished
- fix: name error with 'android.os.Stat**s**Fs'  instead of 'android.os.StatFs' (python names not changed)
- tested on Linux OpenSuSE 42.1
- tested on android (android target) 5.0 (Some notes apply)

### TODO:
- [ ] _Linux_: **processor_info()** returns only processor architecture (for examlple: 'x86_64'),  see https://docs.python.org/3/library/platform.html#platform.processor Maybe we can use **/proc/cpuinfo** for more details?
- [ ] _Android_: **system_name()** returns 'Linux', maybe just return 'Android' instead?
- [ ] _Android_: **version_info()** doesn't break, but try-except excepts.
- [ ] _Android_: **storage_info()** doesn't seem to return neither GBs nor RAM
- [ ] test on _iOS_
- [ ] test on _Windows_
- [ ] test on _macOS_

example screenshot for Linux
kivy_sysinfo_linux


doratoa:
Android screenshot for new changes and fixes (uses modified example app)
screenshot_2016-11-20-20-31-45

doratoa:
All Android issues should be resolved by 5f29e45. Since I didn't get many advices, I've done things in a way I liked.

kiok46:
Hi! @doratoa  android part looks cool!. Thanks you for your work.
one thing.
-  ios, windows and osx return string value in `storage_info` method while android and linux return int.

doratoa:
@kiok46 yes, I know, I didn't work with this platforms yet. There was confusion about storage_info (on android it returned flash storage), so I started remaking it. Meanwhile I decide to return int of bytes instead of formatted string, I believe formatting should be left to user, and we can add optional formatting option in sysinfo facade function so we don't repeat conversion code in each platform

kiok46:
@doratoa  as far as I remember that each platform returned different value for space, for example.
in windows default return value is in `MB` for linux it returns in `Byte` in Mac the default is in `GB`.
We can't just return a integer without informing the user of what that integer's unit is, that's why all the values that are returned should be in one unit.

doratoa:
@kiok46 yeah, I considered a few options, and I'm not 100% sure if my approach is better. At first I though of returning GBs for RAM is more practical, but then I onsidered
1. smaller units may still be sensible for some minimal systems (e.g. Raspberry Pi), 
2. user may want to make comparison for RAM size, thus parsing string with units is not pretty
3. user may want to format with given precision, for once round up 3.84 GB to 4 GB, since platforms seem to return former numbers usually.
4. bytes are most expected unit to return for memory information. Even If it's very unlikely someone really need bytes, if we stop at KB, MB or GB users may ask why exactly that unit. Yes, GB is most used nowadays, but bytes are minimal unit and can be thought as most universal.
After checking that linux and android operate in bytes I concluded that providing integer of bytes everywhere just makes more sense.
We can convert MBs back to bytes (losing some precision, but not critical) - it's kind of stupid, but better than making double conversion everywhere if user is not content with our formatting.
Maybe OSX can report not only GBs, but even if not - we can do the same, since I don't really expect mac with less than 4GB RAM. Won't be that bad.

That's what I thought, but basically I don't mind reverting back to original behaviour.

kiok46:
@doratoa  your points are valid. so we should provide the `Byte` value and let the user decide how they want to convert.
We will provide a method to convert to MB and Gb in the examples.

doratoa:
I'm going to test Windows and add memory/storage changes for it first, then I'll see if I can figure something for Apple platforms.
Nice module, hope we can get it merged soon. I want to use some info in my app too ^_^

doratoa:
Windows revealed to have some things to be fixed too.
First `systeminfo` proved to be horrible source for data. It's somehow slow and what is worse, it's localized.
I've googled up `wmic` command, not sure how reliable it is across Windows variants (available since XP it seems), but it was much more useful.
I also took liberty of replacing version_info with pretty OS name, not sure what to do about returning tuple then, may add extra info like build etc.
Last commit is draft without fixing docs and removing debug parts, but it basically works, screenshot below:
sysinfowin

akshayaurora:
@doratoa are you still working on this ? It would be great to have these opened prs finished and in a mergeable state.

doratoa:
@akshayaurora yeah, I was busy with work up to New Year and then totally went on holidays ^ ^", but have time now and planning to pick up the work. Though my Mac didn't work smoothly last time and most work now is with iOS / macOS, but I hope for better.

Enteleform:
**+1**&nbsp; for access to system info 👍 

Any updates on this?

PR #242, opened 2016-11-14
[WiP] Application dirs

doratoa:
Plyer `appdirs` self-proclaimed analogue for mobile platforms.

Trying to use official guidelines for storing files from Apple and Google
https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html
https://developer.android.com/reference/android/content/Context.html#getCacheDir()
https://developer.android.com/reference/android/content/Context.html#getExternalCacheDir()
https://developer.android.com/reference/android/content/Context.html#getFilesDir()

This code is somewhat tested and should work on Android, iOS and Linux. Android get_cache() could be tricky though.

**TODO**
- [ ] `get_cache()` to get path for large cache files.
- [ ] `get_private()` to get path for files hidden from other apps
- [ ] `get_internal_cache()` to get path on _Android_ for internal small-size cache files (equals `get_cache()` on other platforms)
- [ ] support for Windows (roughly copy `appdirs`)
- [ ] support for macOS (roughly copy `appdirs`)
- [ ] checks for external cache availability on Android
- [ ] get configuration files location
- [ ] point `get_internal_cache()` to `get_cache()` on other platforms
- [ ] test and improve docs


akshayaurora:
@doratoa this seems like a nice feature, got time to finish this ?

Issue #246, opened 2016-12-08
In-app browser


kiok46:
Add support for In-app browser.
Implementation for iOS is here: #194

PR #254, opened 2016-12-26
App browser

phunsukwangdu:
In app browsing implementation for ios and android.
issue #246 
There might be some errors lurking in it thus i request some guidance and support.


plyer/platforms/android/inappbrowser.py
```diff
@@ -0,0 +1,23 @@
+from jnius import cast
+from jnius import autoclass
+from plyer.facades import InAppBrowser
+
+PythonActivity = autoclass('org.renpy.android.PythonActivity')
```
KeyWeeUsr: Should be probably changed to [`org.kivy.android`](https://github.com/kivy/python-for-android/blob/master/pythonforandroid/bootstraps/sdl2/build/src/org/renpy/android/PythonActivity.java#L7), but I'm not sure if there's still some kind of support for pygame or not.

phunsukwangdu: @KeyWeeUsr Here also i have used intent to do stuff but i want to achieve results by using webview of android.For in app browsing  Web view element is there in its  xml files and i could find way to integrate it here.can i have some directions to approach it?

phunsukwangdu:
The ios part of the code is inspired by the pr  #194 which i forgot to mention before.I have used webbrowser to achieve result in linux.The pr is ready.

plyer/platforms/linux/inappbrowser.py
```diff
@@ -0,0 +1,14 @@
+
+from plyer.facades.inappbrowser import InAppBrowser
+import webbrowser
+
+
+class LinuxAppBrowser():
+
+    def _open_url(self, **kwargs):
+        url = kwargs.get('url', '')
```
KeyWeeUsr: Using `''` as the default here is bad because `webbrowser` isn't _only_ for opening the internet browser, but it works even as a file browser, therefore `webbrowser.open('')` even if called from console will probably open Nautilus or similar thing. The same thing works for Windows, where it opens `explorer.exe` in the location you call it from e.g. `%cd%` or `pwd` (maybe?).

phunsukwangdu: Should i will include kivy official site url instead?

KeyWeeUsr: Using `''` as the default here is bad because `webbrowser` isn't _only_ for opening the internet browser, but it works even as a file browser, therefore `webbrowser.open('')` even if called from console will probably open Nautilus or similar thing. The same thing works for Windows, where it opens `explorer.exe` in the location you call it from e.g. `%cd%` or `pwd` (maybe?).

It most likely works even for OSX in the way that such call opens the default browser (like it does for the rest).

Edit: and does the same thing even for Android, so if iOS behaves in the same way, we could probably skip the whole pyobjus thing and replace it with `webbrowser.open('http(s):// ...')`

KeyWeeUsr: and `webbrowser.open('')` will kill the app on android. ^^

dessant: Raising an error if the url is empty might be better.

KeyWeeUsr: Using the right protocol is enough on desktops, but `webbrowser.open('http://')` for some reason kills Firefox on Android. Funny.

phunsukwangdu: Yes it works on linux,windows and osx as stated on [official site](https://docs.python.org/2/library/webbrowser.html) I would love to do it but at present i lack resources to test ios and osx part.So here what i assumed will do in next commit.
- replace ' ' by kivy official site url.
- put the same present linux code under windows and osx.
If there is something more to edit plz suggest :)

phunsukwangdu: ok I will workaround the code to raise error when url is empty.

plyer/platforms/ios/inappbrowser.py
```diff
@@ -0,0 +1,27 @@
...
+UIApplication = autoclass('UIApplication')
+
+
+class iOSInAppBrowser(InAppBrowser):
+    def _open_url(self, **kwargs):
+        url = kwargs.get('url', '')
+
+        nsurl = NSURL.alloc().initWithString_(objc_str(str(url)))
+
+        UIApplication.sharedApplication().openURL_(nsurl)
```
manthansharma: openURL: is deprecated. Use the openURL:options:completionHandler: method instead.

manthansharma:
In OSX there is right implementation for opening InAppBrowser. I think webbrowser.open('') is just opening external browser which doesn't satisfy basics of InAppBrowser.

screen shot 2017-01-22 at 12 20 16 am

plyer/platforms/ios/inappbrowser.py
phunsukwangdu: That is a nice pick.It was deprecated in IOS 10 [link](http://stackoverflow.com/questions/38964264/openurl-in-ios10). Currently i am handicapped due to lack of resources for testing ios part.But soon i will get in touch with my friends who own resources so that i can utilise them to amend changes.

phunsukwangdu:
@manthansharma aspiring gsoc2017 student?

kewok:
I am unable to get this to work using the old_toolchain for python for android.

In particular, the following instructions:

```
# Download the main.py file for appbrowser from examples
git clone https://github.com/phunsukwangdu/plyer.git
cd plyer
git checkout app_browser
cd ..
mkdir TestAppBrowser
cp plyer/examples/inappbrowser/main.py TestAppBrowser/

# Prepare the build environment for python for android
git clone https://github.com/kivy/python-for-android.git
cd python-for-android
git clean -dxf
git pull
git checkout old_toolchain

# Modify the recipe.sh file for plyer to take from branch app_browser in phunsukwangdu/plyer
cp recipes/plyer/recipe.sh recipes/plyer/recipe_BACKUP.sh
sed -i '/VERSION_plyer=${VERSION_plyer:-master}/c\VERSION_plyer=${VERSION_plyer:-app_browser}' recipes/plyer/recipe.sh 
sed -i '/URL_plyer=https:\/\/github.com\/kivy\/plyer\/archive\/$VERSION_plyer.zip/c\URL_plyer=https:\/\/github.com\/phunsukwangdu\/plyer\/archive\/$VERSION_plyer.zip' recipes/plyer/recipe.sh 

# Add android build paths

export ANDROIDSDK=/replace/with/path/to/android-sdk-linux
export ANDROIDNDK=/replace/with/path/to/android-ndk-r10e
export ANDROIDNDKVER=r10e
export ANDROIDAPI=14

# Prepare the build script
./distribute.sh -m "kivy plyer"

cd dist/default

# Build and install the package to my device
./build.py --dir ../../../TestAppBrowser \
    --package org.test.something \
    --name "TestPLYER_INAPP_BROWSER" --version 0.0.1 \
    --orientation sensor \
    --compile-pyo \
    --permission INTERNET debug installd
```

Execute the app on the device, kivy and the inappbrowser interface loads fine. When I hit the Enter button with the default URL, the app crashes and the error appears "Unfortunately, TestPLYER_INAPP_BROWSER has stopped".

Monitoring the output of `adb logcat ` indicates the application dies with
```
03-02 12:56:29.569 24672 24688 I python  :  File "/path_to_python-for-android/installation/python-for-android/build/python-install/lib/python2.7/site-packages/plyer/facades/inappbrowser.py", line 13, in _open_url
03-02 12:24:10.079 18276 18292 I python  :  NotImplementedError
03-02 12:56:29.629 24672 24688 I python  : Python for android ended.
03-02 12:56:29.639 24672 24688 E AndroidRuntime: FATAL EXCEPTION: Thread-13657
03-02 12:56:29.639 24672 24688 E AndroidRuntime: Process: org.test.something:python, PID: 24672
```
However, if I use the `main.py` from `examples/battery` instead (so:` cp plyer/examples/battery/main.py TestAppBrowser/`), and retain the rest of the instructions posted above, I do not get this error and the battery stats display correctly on my android device (Samsung Galaxy Tab A SM-T550).

akshayaurora:
@phunsukwangdu this opens up a browser externally,  in app browser means something else, either change the name of the facade, or fix the implementation

akshayaurora:
@phunsukwangdu please reopen when you have fixed the issues mentioned, this is currently far from mergeable state, the correct implementation would allow the user to easily open a web browser inside the running application itself.

phunsukwangdu:
@akshayaurora yes i will reopen with different name and drop the ios part,right now i am currently preoccupied with kivy text layout system. :)

phunsukwangdu:
@akshayaurora i have changed the title,pr can be reopened.Others working with the projects are free to use my code. :)

Issue #248, opened 2016-12-08
Barcode and QR scanner


kiok46:
Add support for Barcode scanner and QR code scanner.

akshayaurora:
+1

dfrison01:
Tiny bounty added just for fun :)
This topic matters to me. If some pirate has the solution, please leave a message so we can negotiate - sending money via PayPal is very costly since USD isn't the currency in my country.
https://www.bountysource.com/issues/39871127-barcode-and-qr-scanner

Chaostheorie:
@dfrison01 is this still an open issue?

dfrison01:
@Chaostheorie apparently yes. Any help is welcome!

Syfeeq:
@dfrison01 Well, it's been a while since the last reply, not to mention the date this issue was opened, but I've made a simple implementation of QR code (but not only) scanner in Kivy using pyzbar. I have no idea if this is going to be helpful in any way, but I guess it's still better than nothing.

[Kivy-pyzbar](https://github.com/Syfeeq/Kivy-pyzbar)

In case you have any questions then please ask, I'll try to answer them :)

PR #255, opened 2016-12-26
adding barcode/qr scanner functionality

phunsukwangdu:
Qr barcode scanning functionality for  ios and android.I have approached the problem in a naive way
but in future i will be improving it.
issue #248  
There might be some errors lurking in it thus i request some guidance and support. 
Note:- The pr is not in mergeable state  bcz testing is  due and removal of pep8 errors.

### DIFF
--- a/plyer/platforms/android/qrbarcodereader.py
+++ b/plyer/platforms/android/qrbarcodereader.py
@@ -1,0 +1,21 @@
+from jnius import cast
+from jnius import autoclass
+from plyer.facades import QrBarcodeReader
+
+PythonActivity = autoclass('org.renpy.android.PythonActivity')
+Intent = autoclass('android.content.Intent')
+Uri = autoclass('android.net.Uri')
+
+class AndroidQrBarcodeReader(QrBarcodeReader):
+
+    def _scan(self):
+
+        intent = Intent()
+        intent.setAction(Intent.ACTION_VIEW)
+        intent.setData(Uri.parse('http://www.onlinebarcodereader.com'))
+
+        currentActivity = cast('android.app.Activity', PythonActivity.mActivity)
+        currentActivity.startActivity(intent)
+
+def instance():
+    return AndroidQrBarcodeReader()
--- a/README.rst
+++ b/README.rst
@@ -38,6 +38,7 @@ Orientation                        X
 Audio recording                    X
 Flash                              X       X
 Wifi                                           X       X    X
+QrBarcodeScanner                   x       x
 ================================== ======= === ======= ==== =====
 
 Support
--- a/examples/QrBarcodeReader/buildozer.spec
+++ b/examples/QrBarcodeReader/buildozer.spec
@@ -1,0 +1,227 @@
+[app]
+
+# (str) Title of your application
+title = QrBarcode Reader Sample
+
+# (str) Package name
+package.name = QrBarcodeReader
+
+# (str) Package domain (needed for android/ios packaging)
+package.domain = org.test
+
+# (str) Source code where the main.py live
+source.dir = .
+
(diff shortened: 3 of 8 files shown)


plyer/platforms/android/qrbarcodereader.py
```diff
@@ -0,0 +1,21 @@
...
+Intent = autoclass('android.content.Intent')
+Uri = autoclass('android.net.Uri')
+
+class AndroidQrBarcodeReader(QrBarcodeReader):
+
+    def _scan(self):
+
+        intent = Intent()
+        intent.setAction(Intent.ACTION_VIEW)
+        intent.setData(Uri.parse('http://www.onlinebarcodereader.com'))
```
KeyWeeUsr: You can't be serious with that. What if I have tablet and decide to live in a cave for the next month with just a bunch of Qr codes?

phunsukwangdu: Yes it was my naive approach but yes in further commits i am working them out with zxing.

KeyWeeUsr: I'm not sure if [Zxing](https://github.com/zxing/zxing) is the right choice because it's java. And the python thing you could find is only a wrapper (you'll need the java code to build). And that won't be portable on other platforms like... at all this way.

phunsukwangdu: I have seen in tito's repo he has worked out with zbar. [link](https://github.com/tito/android-zbar-qrcode).If not than how should i model this problem?

KeyWeeUsr: Ok then, let's think zxing could work cross-platform. How do you intend to make it work on Apple hardware, which is pretty much ObjC/Swift? You'll still need to compile the Java code somewhere. There surely is a C/Cython library that can handle the things easier. Depending on how good you search it might already have a Python wrapper.

tito: zbar is ok and working on Android. I would be better than nothing, and last time i checked, i didn't find anything that was working crossplatform correctly. An hero may be needed here too, but that's outside the scope here.

IMO, it is ok to have a partial support in plyer, even if there is no direct cross platform solution.

KeyWeeUsr: Wait a sec, Zxing != Zbar as I read it now O_o Zbar alone might [work](https://sourceforge.net/p/zbar/news/2010/09/zbar-iphone-sdk-version-10-released/) even for iphone out-of-box.

KeyWeeUsr: Also, [this](http://www.ama-dev.com/iphone-qr-code-library-ios-7/) might be interesting too. Seems like iphone has something built-in. PyObjus could help probably?

tito: Yep, But anyway, this PR completly depending on a remote website is not good for plyer. Either do a native/library-based implementation, but don't require internet for that or external services :)
+1 for closing and reopen with a new approach.

akshayaurora:
@phunsukwangdu please do a new pr with a better approach using the review/comments as guide, best of luck !

phunsukwangdu:
@akshayaurora I will surely send new pr on this issue.I will act on review/comments to model my new pr.

plyer/platforms/android/qrbarcodereader.py
phunsukwangdu: @tito  i will try native/library-based implementation.It will take little time to implement as i have to research more ways to achieve it on other platforms  but yes  i will do it.I think from ios7 there is inbuilt feature to establish it.

dfrison01:
Any progress on the matter?
Cheers!

PR #259, opened 2017-01-11
add bin and .buildozer directory to .gitignore

malverick:
when we package our app using buildozer then bin and .buildozer directories are created which should not be committed.


akshayaurora:
@malverick thanks

akshayaurora:
2

PR #266, opened 2017-01-19
QRCode Support

manthansharma:
Added QR Code Support for Desktop Environment.

It currently supports scanning code, reading file, encode QR and getting available QR data type list.

screen shot 2017-01-19 at 11 34 34 pm

Supported OS
- [x] OSX
- [x] Linux 
- [ ] Windows

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,7 +35,7 @@ Gyroscope                          X       X
 Battery                            X       X   X       X    X
 Native file chooser                            X       X    X
 Orientation                        X
-QR Code
+QR Code                                                X    X
 Audio recording                    X
 Flash                              X       X
 Wifi                                           X       X    X
--- a/README.rst
+++ b/README.rst
@@ -35,6 +35,7 @@ Gyroscope                          X       X
 Battery                            X       X   X       X    X
 Native file chooser                            X       X    X
 Orientation                        X
+QR Code
 Audio recording                    X
 Flash                              X       X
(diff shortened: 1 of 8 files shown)


KeyWeeUsr:
Unfortunately this approach uses [`qrtools`](https://github.com/primetang/qrtools) which is GPL 3.0 and internally [ZBar](http://zbar.sourceforge.net/) which is LGPL 2.1. Both can't be sublicensed, therefore with `qrtools` all of the code anyone would use would need to be licensed under GPL 3.0 (or higher depending on the exact license terms), which is a no go + at least one big mobile application store forbids such license.

With ZBar alone we might be able to use it, but we still might be on a thin ice, so I'd rather see something that wouldn't cause legal issues for anyone packaging plyer (and its dependencies) into single archive/binary.

AndreMiras:
Good catch @KeyWeeUsr but pretty sad to see nice feature not being part of the tree for licensing stuff.
I'm wondering to what extent we could simplify integration with incompatible license without risking kivy.

KeyWeeUsr:
@AndreMiras The feature yes, the approach no. Re the licensing stuff, we can't IMO. If it's not MIT compatible (plyer and most of Kivy org repos' licenses) we can't really do much unless we want to cause damage to the people using our libraries.

As I wrote, we might be able to use ZBar in case someone implements it with that, but I'd rather see something with less restrictive license as a package we should depend upon.

PR #267, opened 2017-01-19
Multilingual Support for TextToSpeech

manthansharma:
Added new method `language()` for getting all supported language. 

Tested on OSX, Linux, MacOSX.

- [ ]  Update Example
- [x]  Desktop Support

PR #268, opened 2017-01-21
Speech to Text Support (WIP)

manthansharma:
Added speech to text support on all desktop environment and tested. Implemented using Google Speech Recognition API. Audio source can be file or microphone.

- [x] Desktop Env.
- [ ] Example

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,8 @@ Call                               X       X
 Camera (taking picture)            X       X
 GPS                                X       X
 Notifications                      X           X       X    X
-Text to speech                     X       X   X       X    X
+Text to Speech                     X       X   X       X    X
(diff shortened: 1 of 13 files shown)


dessant:
Thanks for your work, but it does not look like a good fit for plyer. We do not want to use a paid web api, when OS apis are available.

https://developer.apple.com/reference/appkit/nsspeechrecognizer
https://msdn.microsoft.com/en-us/library/jj127860.aspx

If your tts espeak changes correct and issue, please submit them in a new pr, explaining what is being fixed.

manthansharma:
Ok, i have submitted TTS patch in diff. PR #267. I will try to use OS API for implementation.

Issue #3, opened 2014-01-21
Speech to Text


brousch:
A possible solution for cross-platform (desktops) Speech to Text http://cmusphinx.sourceforge.net/

DebVortex:
@trivedigaurav I may be a little bit of a "grave digger" here by writing in this old and closed Issue, yet isn't the reason why you closed it wrong?

The purpose of the Issue was bringing the Speech to Text (STT) feature into plyer. However, you closed it with the argumentation, "Text to Speech" (TTS) is now implemented.

Sorry, but did I miss something? Isn't the Issue meant "the other way around"?

trivedigaurav:
Oops. You are right. This was probably in error. Let me find if there is an open TTS issue somewhere.

eadmaster:
any update on android support?
As alternative I've found it is supported by [SL4A](https://github.com/damonkohler/sl4a/blob/master/python/ase/scripts/speak.py).

DebVortex:
@eadmaster both proposed links take care of text to speech, which is supported by plyer. However, this Issue targets the other way around: speech to text.

eadmaster:
oops, sorry you're right!

Blind0ne:
Is there a way now to use speech recognition in kivy on android? With libraries like https://pypi.org/project/SpeechRecognition/ ?

KeyWeeUsr:
Judging by the dependencies, contents of the repository (**binaries** without license, 3rd party modules included, etc etc) and [implementation](https://github.com/Uberi/speech_recognition/blob/master/speech_recognition/__init__.py) `SpeechRecognition` package has, no. Nothing like that.

That package basically first records an audio from microphone, then dumps it to a file and sends the file to Google API via an HTTP request.

The point of plyer isn't being a wrapper for a wrapper for a wrapper ... for an API, nor it is practical in case on of the dependencies either disappears of has a bug or does not compile correctly and more of those issues. And this one is even worse because it even uses a web API which for you as a library user means requesting not only audio permissions, but networking permissions too.

From the licensing side unless you have all the licenses, you provide or sell your packaged software (containing dependencies) and don't include the licenses and fulfill the demands of each license you are violating those licenses and the copyright for each such a piece of dependency.

[...]

PR #269, opened 2017-01-21
Speech to Text (WIP)

manthansharma:
#267 As Google Speech Recognition API is paid version moving to [CMU Sphinx](http://cmusphinx.sourceforge.net/wiki/), which is free offline version for Speech Recognition.

Added speech to text support on all desktop environment and tested. Implemented using [CMU Sphinx.](http://cmusphinx.sourceforge.net/wiki/) Audio source can be file or microphone.

- [x] Desktop Env.
- [ ] Example

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,7 @@ GPS                                X       X
 IR Blaster                         X
 Notifications                      X           X       X    X
 Text to Speech                     X       X   X       X    X
-Speech to Text                                         X
+Speech to Text                                 X       X    X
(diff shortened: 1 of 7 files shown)


manthansharma:
Windows part can be implemented using https://msdn.microsoft.com/en-us/library/jj127860.aspx. is this module is good or have to use OS apis.

KeyWeeUsr:
ref https://github.com/kivy/plyer/issues/3#issuecomment-437608385

Issue #261, opened 2017-01-13
Send SMS feature not working


malverick:
When I press the 'Send Sms' button then I receive the warning that 'charges may apply' and after agreeing to the warning nothing happens.

malverick:
I am working on the fixing the issue and will send the pr once done.

akshayaurora:
did you check if you gave the app the appropriate permissions? Just tested with 10 + devices in a workshop, it works.

shahzadakram67:
When I press the 'Send Sms' button then my app crashes why is this so,,???

PR #271, opened 2017-01-22
Send sms android api fix

malverick:
Fixes issue #261 
Current android sms api in plyer uses sms manager api to send sms. I have tried to use intent class which uses built-in sms app of device to send sms.
Any suggestions are welcomed otherwise it is ready to be merged.

### DIFF
--- a/examples/sms/buildozer.spec
+++ b/examples/sms/buildozer.spec
@@ -32,7 +32,7 @@ source.include_exts = py,png,jpg,kv,atlas
 version = 0.1
 
 # (list) Application requirements
-requirements = plyer,kivy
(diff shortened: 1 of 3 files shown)


malverick:
screenshot_20170123-053603

After pressing `Send SMS` button,

screenshot_20170123-053630

akshayaurora:
Original implementation works, just tested it yesterday on multiple mobiles in a work shop 10+ mobiles with 0 issues.

Feel free to open a new one, adding option `launch_interface` defaulting to 0 preserving old behavior

akshayaurora:
@malverick thanks for all the work :)

PR #273, opened 2017-01-26
orientation feature for linux

phunsukwangdu:
Orientation feature for linux. It has been implemented with the help of xrandr.
Pr is in working condition and ready to be merged.

### DIFF
--- a/plyer/platforms/linux/orientation.py
+++ b/plyer/platforms/linux/orientation.py
@@ -7,17 +7,17 @@ class AndroidOrientation(Orientation):
     def _set_landscape(self, **kwargs):
         self.rotate = 'normal'
         self.screen = sb.check_output("xrandr -q | grep ' connected' \
-                              |  head -n 1 | cut -d ' ' -f1", shell = True)
+                              |  head -n 1 | cut -d ' ' -f1", shell=True)
         self.screen = self.screen.split('\n')[0]
         sb.call(["xrandr", "--output", self.screen, "--rotate", self.rotate])
 
     def _set_portrait(self, **kwargs):
         self.rotate = 'left'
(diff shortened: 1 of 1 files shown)


plyer/platforms/linux/orientation.py
```diff
@@ -0,0 +1,23 @@
+import subprocess as sb
+from plyer.facades import Orientation
+
+
+class AndroidOrientation(Orientation):
```
malverick: I think it should be LinuxOrientation

malverick: I think it should be LinuxOrientation

plyer/platforms/linux/orientation.py
```diff
@@ -0,0 +1,23 @@
...
+    def _set_portrait(self, **kwargs):
+        self.rotate = 'left'
+        self.screen = sb.check_output("xrandr -q | grep ' connected' \
+                              |  head -n 1 | cut -d ' ' -f1", shell=True)
+        self.screen = self.screen.split('\n')[0]
+        sb.call(["xrandr", "--output", self.screen, "--rotate", self.rotate])
+
+
+def instance():
+    return AndroidOrientation()
```
malverick: return LinuxOrientation()

phunsukwangdu:
thx malverick i was doing the changes meanwhile you spotted it. :)

akshayaurora:
@phunsukwangdu thanks 👍 could you add a try/except and throw a appropriate error just to exit cleanly when Xrandr is not installed.

bhaveshAn:
Using `shell=True` is hazardous Please see https://docs.python.org/2/library/subprocess.html#frequently-used-arguments

PR #274, opened 2017-02-02
Add on_pause function

malverick:
Some of the example apps does not contain `on_pause` function due to which, when the app is relaunched after pressing the home button, app restarts. Adding the `on_pause` function will cause the app to start from state where it was before leaving.

### DIFF
--- a/examples/accelerometer/basic/main.py
+++ b/examples/accelerometer/basic/main.py
@@ -51,6 +51,6 @@ class AccelerometerTestApp(App):
 
     def on_pause(self):
         return True
-        
+
 if __name__ == '__main__':
     AccelerometerTestApp().run()
--- a/examples/accelerometer/basic/main.py
+++ b/examples/accelerometer/basic/main.py
@@ -49,5 +49,8 @@ class AccelerometerTestApp(App):
     def build(self):
         return AccelerometerTest()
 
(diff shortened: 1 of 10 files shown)


kiok46:
@malverick  Thanks.

dessant:
@kiok46, @malverick, this is becoming `unnecessary` since on_pause returns True in kivy master.

kiok46:
kivy/kivy#4603 Now I remember, thanks.

malverick:
Hi @rockcastle . Can you please share the android logs?

rockcastle:
Hi, 
Sorry, I already clear logs, I can send debugged apk if it is ok.

malverick:
That would be fine.

rockcastle:
I have tried one more time creating apk and done also it works fine on tablet which is android 5.1.          I have uploaded my buildozer.spec and main.py to my GitHub ( https://github.com/rockcastle/Python-Kivy-Android-app )
I have realized that autoclass('org.rendy.android.PythonActivity') not used any on crystax-ndk python3, instead of use autoclave('org.kivy.android.PythonActivity')

By the way thanks.

dessant:
@rockcastle, please move the discussion to irc or the mailing list.

PR #275, opened 2017-02-05
Linux camera feature

phunsukwangdu:
The feature captures photos  and videos for linux.It is implemented with the help of gstreamer-tools.
To stop video recording one needs to press cltr+c.
[](https://postimg.org/image/e6cao9vh3/)


akshayaurora:
what happens when gat-launch does not exist on a system?

akshayaurora:
Assuming gstreamer would be available on most linux systems would be a stretch, please look at using more generic methods like using v4l2 which has a larger penetration or something equivalent that would be guaranteed to have the binaries in the linux world.

phunsukwangdu:
hi @akshayaurora i am currently not focusing on plyer.I am currently working to integrate harfbuzz shaper into kivy.As soon i  do that i will focus on plyer,meanwhile @bhaveshAn has leads on it :).

matham:
@phunsukwangdu It is not in the spirit of OSS (which is all about cooperation so that we can make the best software in the best environment) and it is certainly not acceptable for us to make such condescending and territorial remarks on another person's issue. Whether they are a gsoc aspirant or not. Please refrain from doing that in the future.

matham:
@phunsukwangdu sorry, about me previous comment, the PRs looked far too similar to bhaveshAn's on this subject, but I guess you're was actually first.

PR #276, opened 2017-02-06
Error catch addition

phunsukwangdu:
Error popup for examples where feature implementation(NotImplementedError) doesn't take place on the platform.All are examples are tested and are in merge able state.

### DIFF
--- a/examples/call/main.py
+++ b/examples/call/main.py
@@ -4,6 +4,8 @@ from kivy.uix.button import Button
 from kivy.lang import Builder
 from kivy.properties import StringProperty
 from plyer import call
+from kivy.uix.popup import Popup
+from kivy.uix.label import Label
 
 Builder.load_string('''
 #: import Platform kivy.utils.platform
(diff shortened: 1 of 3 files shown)


examples/flash/main.py
```diff
@@ -24,13 +26,34 @@
...
     def release(self):
-        flash.release()
+        try:
+            flash.release()
+        except NotImplementedError:
+            self.ErMsg = "Feature under development for this platform!"
+            popup = Popup(title="Error!",
+                          content=Label(text=self.ErMsg),
+                          size_hint=(None, None), size=(350, 350))
+            popup.open()
```
kiok46: Could these be done using one method instead of doing it 3 time?
and maybe if self.ErMsg says "Feature not available/implemented for this platform." that would make more sense?

examples/call/main.py
```diff
@@ -41,14 +43,28 @@ class CallInterface(BoxLayout):
...
     def call(self, *args):
-        call.makecall(tel=self.tel)
+        try:
+            call.makecall(tel=self.tel)
+        except NotImplementedError:
+            self.ErMsg = "Feature under development for this platform!"
+            popup = Popup(title="Error!",
+                          content=Label(text=self.ErMsg),
+                          size_hint=(None, None), size=(350, 350))
+            popup.open()
```
kiok46: same.

examples/flash/main.py
phunsukwangdu: @kiok46 I will reflect the following changes according to your feeds in my next commit. :)

* Make a separate class for popup so that it is not implemented three time separately and reflect the same to other files.
* Change error message to "Feature not implemented for this platform!"

PR #278, opened 2017-02-09
Android Gravity Sensor

malverick:
@dessant I have tested the api and done the required changes(much changes were not required though).
I request you to review the pr. From my side, it can be merged.
screenshot_2017-02-09-18-18-15-137_org test plyergravity
screenshot_2017-02-09-18-18-25-355_org test plyergravity

PR #279, opened 2017-02-09
Android gravity sensor

malverick:
Gravity Sensor api for android platform.
Tested on:
Device - Xiaomi Mi 4i
Toolchain - old and new both
Android api - 19
Android sdk - 24
screenshot_2017-02-09-18-18-15-137_org test plyergravity
screenshot_2017-02-09-18-18-25-355_org test plyergravity

@dessant now you can review it.

### DIFF
--- a/plyer/platforms/android/gravity.py
+++ b/plyer/platforms/android/gravity.py
@@ -64,7 +64,7 @@ class AndroidGravity(Gravity):
             self.listener.disable()
             del self.listener
 
-    def _get_magnetic(self):
+    def _get_gravity(self):
         if self.listener:
             return tuple(self.listener.values)
         else:
--- a/plyer/platforms/android/gravity.py
+++ b/plyer/platforms/android/gravity.py
@@ -52,26 +52,30 @@ class GravitySensorListener(PythonJavaClass):
 
 class AndroidGravity(Gravity):
 
-    listener = None
+    def __init__(self):
+        super(AndroidGravity, self).__init__()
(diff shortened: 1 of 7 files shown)


plyer/platforms/android/gravity.py
```diff
@@ -52,26 +52,30 @@ def onAccuracyChanged(self, sensor, accuracy):
...
 class AndroidGravity(Gravity):
 
-    listener = None
+    def __init__(self):
+        super(AndroidGravity, self).__init__()
+        self.state = False
 
     def _enable(self):
-        if not self.listener:
+        if (not self.state):
```
dessant: What is the reason for adding the brackets?

malverick: it will work fine without brackets. Its just my habit of putting brackets with if condition. It looks more simplified to me. Does it violate any rule of coding style?

dessant: Yes, it violates the unwritten rule of not adding stuff that is not necessary. 😀

malverick: ok I will do the changes and will remember it from next time.

dessant:
@malverick, please be very attentive about our requests, 'Remove unnecessary stuff' is not a good commit message, and the pr description does not mention building on the original pr. These were all discussed before: https://github.com/kivy/plyer/pull/278.

When additional changes are made, the pr must explain what is the reason for the change, and what issue is being fixed.

PR #280, opened 2017-02-09
Plyer android gravity sensor

malverick:
Gravity sensor api for android platform.
My work is based on pr #161 and I have done some additional changes to bring it in merge-able state.

In file `plyer/platforms/android/gravity.py` in class `AndroidGravity` an object `listener` of class `GravitySensorListener()` was used to check whether sensor is active/inactive and the same was used to activate/deactivate the sensor. I think that was causing an issue because when `_get_gravity(self)` function was called, `self.listener.values` was returning `(None,None,None)`.
img_20170210_001548_334
img_20170210_001529_089

Instead, I defined a constructor in which I am using a different variable `state` to check whether sensor is active/inactive and `listener` object is used for enabling/disabling the sensor and for readings too as before.
img_20170210_001254_497
img_20170210_001321_250

Rebasing is also done since the original pr was 114 commits behind the master.

Tested on:
Device - Xiaomi Mi 4i
Toolchain - old and new both
Android api - 19
Android sdk - 24

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,7 +44,6 @@ Wifi                                           X       X    X
 Support
 -------
 
-<<<<<<< 882a7ca7209318c9f9b5f4684e4f8c5add8754f1
 If you need assistance, you can ask for help on our mailing list:
 
 * User Group : https://groups.google.com/group/kivy-users
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,7 @@ Accelerometer                      X       X           X    X
 Call                               X       X
 Camera (taking picture)            X       X
 GPS                                X       X
+Gravity                            X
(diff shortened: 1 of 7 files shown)


dessant:
Ok, let's do one more round, the third attempt will likely have more explicit commit messages, and not contain unnecessary commits, like https://github.com/kivy/plyer/pull/280/commits/9682f2d398767c2801bfc9668312b5773ff60e2f, since we already established that the brackets are not needed, before this last pr was made, over at https://github.com/kivy/plyer/pull/279#discussion_r100337800). `update readme` looks like it solves a merge conflict, it should say so.

The pr description does look OK. 👍

PR #281, opened 2017-02-09
Add Keystore Functionality

Kovak:
Depends on the[ keyring](https://pypi.python.org/pypi/keyring) module for desktop implementations. Is that inline with plyer's goals?


Kovak:
actually this need some api fixes and documentation still

PR #283, opened 2017-02-10
Plyer android gravity sensor

malverick:
Gravity sensor api for android platform.
My work is based on pr #161 and I have done some additional changes to bring it in merge-able state.

In file `plyer/platforms/android/gravity.py` in class `AndroidGravity` an object `listener` of class `GravitySensorListener()` was used to check whether sensor is active/inactive and the same was used to activate/deactivate the sensor. I think that was causing an issue because when `_get_gravity(self)` function was called, `self.listener.values` was returning `(None,None,None)`.
img_20170210_001548_334
img_20170210_001529_089

Instead, I defined a constructor in which I am using a different variable `state` to check whether sensor is active/inactive and `listener` object is used for enabling/disabling the sensor and for readings too as before.
img_20170210_001254_497
img_20170210_001321_250

Rebasing is also done since the original pr was 114 commits behind the master.

Tested on:
Device - Xiaomi Mi 4i
Toolchain - old and new both
Android api - 19
Android sdk - 24


malverick:
Thank you @dessant

PR #284, opened 2017-02-10
Android app navigation from notifications

nlopez73:


Fogapod:
Notification is disappearing after returning to app using notification. I think this shouldn't work this way

nlopez73:
I'm not sure, but I think notification is disappearing after using notification, it doesn't mind if you returning to app or not.

PR #159, opened 2015-06-12
Plyer android proximity

thegrymek:
Hi,
Here is pull request with proximity Facade with example and android implementaion.

best regards,
Andrzej Grymkowski

PR #287, opened 2017-02-13
Plyer android proximity sensor

malverick:
Proximity sensor api for android platform.

The work is based on #159 and I have done some additional changes to bring it in merge-able state.

**Issue in the original pr** - Once the proximity sensor is enabled, after covering the sensor, it should display `yes` meaning `sensor has detected a object` instead it was always displaying `no` irrespective of whether it is covered or not. Maybe this was due to not returning correct value by `_get_proximity` function.

I made the required changes and after that, the example is working something like this - 
**Initially**

screenshot_2017-02-14-00-20-08-144_org test plyer proximity

**Just after pressing `enable`**

screenshot_2017-02-14-00-20-18-707_org test plyer proximity

**After covering the top part of mobile where sensor is located**

screenshot_2017-02-14-00-20-46-977_org test plyer proximity

Additionally, rebasing is also done since the original pr is `117` commits behind `master`.

**Tested on** -
**Devices** - Xiaomi Mi 4i, Samsung Galaxy S6 Edge
**Toolchain** - old and new both
**Android api** - 19
**Android sdk** - 24


plyer/platforms/android/proximity.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def _enable(self, **kwargs):
+        if not self.listener:
+            self.listener = ProximitySensorListener()
+            self.listener.enable()
+
+    def _disable(self, **kwargs):
+        if self.listener:
+            self.listener.disable()
+            delattr(self, 'listener')
```
dessant: I think this will crash if the sensor is enabled and disabled then either enabled or disabled again.

```python
class Test():
    prop = None


var = Test()
delattr(var, 'prop')
print(var.prop)
```

It may be better to just use `self.listener = None`.

dessant: Scratch that, i've made a mistake. 😋 

Of course it works:

```python
class Test():
    prop = None

    def set(self):
        self.prop = 1

    def delete(self):
        delattr(self, 'prop')


var = Test()
var.set()
var.delete()

print(var.prop)
```

malverick:
Thanks @dessant

dessant:
@malverick, thanks for the fixes. 👍

PR #291, opened 2017-02-22
Android Spatial orientation sensor

malverick:
Spatial orientation sensor api for android platform.
My work is based on pr #157 and I have done some additional changes to bring it in a merge-able state.

The original pr implemented screen orientation along with spatial orientation but since the screen orientation part is already implemented in plyer so I have kept only the spatial orientation part. Since, screen and spatial orientation are conceptually different therefore I have created a new api instead of updating the old one.

In addition to this, rebasing is also done since the original pr branch is 125 commits behind master branch.
screenshot_2017-02-23-04-26-23-321_org test plyerspatialorientation
screenshot_2017-02-23-04-26-29-815_org test plyerspatialorientation

Tested on -
Device: Xiaomi Mi 4i
Toolchain - old and new both
Android api - 19
Android sdk - 24

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ venv
 
 bin
 .buildozer
+.idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,939 +1,216 @@
-Changelog
-=========
-
-%%version%% (unreleased)
-------------------------
-
-- Update __init__.py. [dessant]
-
-- Update copyright year. [dessant]
-
-- Fix windows filechooser using `mode` instead of `self.mode` [gabriel
-  pettier]
-
-- Merge pull request #144 from thegrymek/plyer-audio. [dessant]
-
-  Plyer audio for android with facade and example
-
-- Update docs facade with permissions. [andrzej.grymkowski]
-
-- Improving docs. [andrzej.grymkowski]
-
-- Corrected "" in docstring. [Andrzej Grymkowski]
-
-- Flake8. [Andrzej Grymkowski]
-
-- Update readme. [Andrzej Grymkowski]
-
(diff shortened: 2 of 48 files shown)


malverick:
The sensor used here, `TYPE_ORIENTATION` was deprecated in API Level 8.
An alternate way to get the values of `azimuth`, `pitch` and `roll` would be to use `SensorManager.getOrientation()` method. [Reference](https://developer.android.com/reference/android/hardware/SensorManager.html#getOrientation(float[],%20float[]))
I will update the code accordingly.

malverick:
Now, I have derived the orientation sensor reading using `rotation matrix`.
Sensors used for calculating the readings are `TYPE_ACCELEROMETER` and `TYPE_MAGNETOMETER`
API is tested and working fine and can be merged.
**Tested on:**
**Device -** Xiaomi Mi 4i
**Android sdk -** 24
**Android ndk -** r9c
**Androd api -** 19
index

bhaveshAn:
@malverick These are the results of Magnetometer only and there's a PR concerning this #300 which is also tested by you few days back. I am not getting what are you trying to do then. If any specific reason please explain with some facts.

malverick:
@bhaveshAn this pr focuses on implementation of `spatial orientation` sensor which gives the `orientation vector` of device in form of `azimuth`, `pitch` and `roll` based on `rotation matrix` calculated using `magnetometer`. The android API for this type of sensor is deprecated therefore I have used `magnetometer` and `accelerometer` to derive the results. For more information, you can go through the reference link shared above.
The unit of readings in `magnetometer` is ` μT` and here is `rad/s`. How can they be the same?

malverick:
closed in favor of #336

PR #293, opened 2017-02-28
Android ambient temperature sensor

malverick:
Android temperature sensor api to measure the ambient room temperature in degrees Celsius (°C).
Can be merged.

Example for this api is not tested since I do not have access to android device having temperature sensor. It would be very helpful if someone can test it.

P.S. -  I am sure it will run successfully because it is almost similar to light and pressure sensor api and they are working fine.


malverick:
Thanks @dessant

PR #294, opened 2017-03-01
Linear acceleration

bhaveshAn:
Android Linear Acceleration:  Acceleration forces along the axes (excluding gravity).
Previously, 
Android Accelerometer:  Acceleration forces along the axes (including gravity).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,6 +38,7 @@ Orientation                        X
 Audio recording                    X
 Flash                              X       X
 Wifi                                           X       X    X
+Linear Acceleration                X
 ================================== ======= === ======= ==== =====
 
 Support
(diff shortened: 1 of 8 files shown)


bhaveshAn:
Try using
Android sdk - 24
Android api - 19

KeyWeeUsr:
I don't think this should be separated from `accelerometer` at all. Instead this should be part of the already implemented API. Currently we have `accelerometer.acceleration` for relative one and we could switch that into more parts:

* `accelerometer.raw`
* `accelerometer.linear`
* etc

What all of those in Android do is [explained here](https://stackoverflow.com/a/8006832/5994041). How to do the same on iOS is [here](https://stackoverflow.com/a/42209192/5994041). Unless there is implementation for iOS it makes the API inconsistent though.

PR #297, opened 2017-03-10
fix handling of notifications' timeout on Linux

benoit-pierre:
- the API expects a value in seconds (but the DBus API uses milliseconds)
- the documentation says the default timeout is 10 seconds (not 5)


kiok46:
Thanks :)

PR #298, opened 2017-03-10
Android rotation vector

bhaveshAn:
Android Rotation Vector giving following values:

Rotation vector component along the x-axis (x * sin(θ/2))
Rotation vector component along the y-axis (y * sin(θ/2))
Rotation vector component along the z-axis (z * sin(θ/2))
Scalar component of the rotation vector ((cos(θ/2))

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,6 +38,7 @@ Orientation                        X
 Audio recording                    X
 Flash                              X       X
 Wifi                                           X       X    X
+RotationVector                     X
 ================================== ======= === ======= ==== =====
 
 Support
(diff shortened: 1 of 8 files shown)


bhaveshAn:
Try using
Android sdk - 24
Android api - 19

bhaveshAn:
Compiling and tested. Although not have access to device having rotation vector sensor. requesting to test this on some device and reply either way.

ANDROID-SDK : 24
ANDROID-NDK : r9c
ANDROID-API : 19

rotation before

rotation after

KeyWeeUsr:
Same as https://github.com/kivy/plyer/pull/294#issuecomment-431619930 since this is a part of the accelerometer data, so it shouldn't be separated. Until there is no [iOS alternative](https://stackoverflow.com/a/24958690/5994041) it still introduces inconsistent API though.

PR #158, opened 2015-06-12
Plyer android stepcounter

thegrymek:
Example with android implementation.

PR #299, opened 2017-03-10
Android Step Counter feature added

bhaveshAn:
Step Counter sensor returns the number of steps taken by the user since the last reboot while activated. The value is returned as a float (with the fractional part set to zero) and is reset to zero only on a system reboot. 

If someone wants to continuously track the number of steps over a long period of time, do NOT unregister for this sensor, so that it keeps counting steps in the background even when the AP is in suspend mode and report the aggregate count when the AP is awake. Application needs to stay registered for this sensor because step counter does not count steps if it is not activated

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,7 +37,8 @@ Native file chooser                            X       X    X
 Orientation                        X
 Audio recording                    X
 Flash                              X       X
-Wifi                                           X       X    X
+Wifi                                           X       X    X           
+Step Counter                       X
 ================================== ======= === ======= ==== =====
 
 Support
--- a/examples/stepcounter/buildozer.spec
+++ b/examples/stepcounter/buildozer.spec
@@ -1,0 +1,227 @@
+[app]
+
+# (str) Title of your application
+title = PlyerStepCounter
+
+# (str) Package name
+package.name = myapp
+
+# (str) Package domain (needed for android/ios packaging)
+package.domain = org.test
+
+# (str) Source code where the main.py live
+source.dir = .
+
+# (list) Source files to include (let empty to include all the files)
+source.include_exts = py,png,jpg,kv,atlas
+
+# (list) List of inclusions using pattern matching
+#source.include_patterns = assets/*,images/*.png
+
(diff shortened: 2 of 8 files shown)


bhaveshAn:
Try using
Android sdk - 24
Android api - 19

bhaveshAn:
SENSOR_DELAY_FASTEST  used to get sensor data as fast as possible 
SENSOR_DELAY_NORMAL  used for rate (default) suitable for screen orientation changes 
SENSOR_DELAY_UI  used for rate suitable for the user interface 

#158 Here Sensor is registered as SENSOR_DELAY_UI but SENSOR_DELAY_FASTEST should be used as step counter is enabled as SENSOR_DELAY_FASTEST 
please see examples http://stackoverflow.com/questions/38650354/how-to-reset-the-step-count-from-android
http://stackoverflow.com/questions/23520814/simple-android-pedometer-not-working

dessant:
Please rename the facade to `pedometer`.

README.rst
```diff
@@ -42,6 +42,8 @@ Orientation                        X
 Proximity                          X
 Audio recording                    X
 Flash                              X       X
+Wifi                                           X       X    X           
+Pedometer                       X
```
malverick: make it aligned to above ones

bhaveshAn:
Compiling and tested. Although not have access to device having relative humidity sensor. requesting to test this on some device and reply either way.

pedometer before


pedometer after

tomplast:
I just tested this feature from bhaveshAn's repository and it worked fine for me, I just had to had ACTIVITY_RECOGNITION to permissions in buildozer.spec

tomplast:
@akshayaurora Could you please take a look at this?

misl6:
Hi @tomplast ,

Seems we have some conflicts that need to be resolved before merging.

Since you recently tested the feature, and seems to work, what do you think about proposing a new PR in alternative to @bhaveshAn one (the PR has been opened in 2017, and I do not know if the original author still has the time to contribute).

Please include usage instructions (permissions, etc) in docs and a demo app . 😀

PR #301, opened 2017-03-10
Android Humidity

bhaveshAn:
tells the humidity in percentage


bhaveshAn:
Try using 
Android sdk - 24
Android api - 19

malverick:
Tested and compiling fine. Can't say about the working of it since I don't have access to any device having relative humidity sensor.
humidity

bhaveshAn:
Compiling and tested. Although not have access to device having relative humidity sensor. requesting to test this on some device and reply either way.

relative humidity before

relative humidity after

bhaveshAn:
Since the relative humidity is not the kind of sensor available on other platforms. Instead changed the implementation from Relative Humidity sensor to Humidity which is the absolute humidity using the two sensors `TYPE_RELATIVE_HUMIDITY` and `TYPE_AMBIENT_TEMPERATURE` .
Formulae used to get absolute humidity using reading of these two sensors is available [here](https://developer.android.com/guide/topics/sensors/sensors_environment.html) .

` humidity = (Ta * (Rh / 100) * A * exp(m * Tc / (Tn + Tc)) /(K + Tc))`

where,

    Tc = actual temperature in degrees C
    Rh = actual relative humidity in percent (%)
    m = 17.62
    Tn = 243.12 degrees C
    A = 6.112 hPa
    K = 273.15

akshayaurora:
@bhaveshAn thanks

morgansong:
Hi, Mr. I had a Samsung S4 with Sensirion humidity and temperature sensor. I try the buildozer.spec and main.py to try. For temperature, it is ok to show the value but for relative humidity, the compiling process sometime crash or sometime, can't install the apk, or sometime can install but will close automatically very fast. I use the Buildozer VM in Windows to build the apk.

PR #303, opened 2017-03-12
Android Game rotation vector sensor

malverick:
Game Rotation Vector sensor api for android platform.
A constant describing an uncalibrated rotation vector sensor type. 

P.S. - Feature is not yet tested and I will add the screenshots after testing the app within few hours along with the device information.


malverick:
Tested and working fine.
Device - Mi Redmi Note 3
Android sdk - 24
Android ndk - r9c
Android api - 19
game
game1

malverick:
This type of sensor is available only for android platform and after digging into the apis provided by windows and apple, I couldn't find any equivalent api for this type of sensor and also there doesn't exist any way of extracting this type of sensor data from existing sensors. 
It anyone finds a possible way of implementing a equivalent api in any other platform, please post here.
Thank you
References - 
https://msdn.microsoft.com/en-us/library/windows/desktop/dd318969(v=vs.85).aspx
https://developer.apple.com/reference/coremotion

PR #304, opened 2017-03-12
Android Geomagnetic rotation vector sensor

malverick:
Geomagnetic rotation vector sensor api for android platform.
The geomagnetic rotation vector sensor is similar to the Rotation Vector Sensor, but it uses a magnetometer instead of a gyroscope. The accuracy of this sensor is lower than the normal rotation vector sensor, but the power consumption is reduced. Only use this sensor if you want to collect some rotation information in the background without draining too much battery. This sensor is most useful when used in conjunction with batching.

P.S. - Feature is not yet tested and I will add the screenshots after testing the app within few hours along with the device information.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,6 +39,7 @@ Orientation                        X
 Audio recording                    X
 Flash                              X       X
 Wifi                                           X       X    X
+Geomagnetic Rotation Vector        X
 ================================== ======= === ======= ==== =====
 
 Support
--- a/examples/geomagneticrotationvector/buildozer.spec
+++ b/examples/geomagneticrotationvector/buildozer.spec
@@ -1,0 +1,233 @@
+[app]
+
+# (str) Title of your application
+title = Geomagnetic Rotation Vector Example
+
+# (str) Package name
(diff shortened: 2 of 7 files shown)


malverick:
Tested and working fine.
Device - Mi Redmi Note 3
Android sdk - 24
Android ndk - r9c
Android api - 19
geo
geo1

malverick:
This type of sensor is available only for android platform and after digging into the apis provided by windows and apple, I couldn't find any equivalent api for this type of sensor and also there doesn't exist any way of extracting this type of sensor data from existing sensors. 
It anyone finds a possible way of implementing a equivalent api in any other platform, please post here.
Thank you
References - 
https://msdn.microsoft.com/en-us/library/windows/desktop/dd318969(v=vs.85).aspx
https://developer.apple.com/reference/coremotion

KeyWeeUsr:
Since this is Android specific only, as you describe, I'm not sure it needs to be implemented, since there is `pyjnius` for stuff like that. Anyway, the end result should be included under `magnetometer` as it uses the same sensor.

PR #311, opened 2017-03-18
Linux Audio feature added

bhaveshAn:
Linux audio feature is added and it doesn't having any dependencies. Instead, feature is running using gstreamer only.

Please view this PR.

### DIFF
--- a/plyer/facades/temperature.py
+++ b/plyer/facades/temperature.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 
+
 class Temperature(object):
     '''Temperature facade.
 
@@ -25,7 +26,7 @@ class Temperature(object):
         '''Disable temperature sensor.'''
         self._disable()
 
-    #private
+    # private
 
     def _get_temperature(self, **kwargs):
         raise NotImplementedError()
--- a/plyer/facades/temperature.py
+++ b/plyer/facades/temperature.py
@@ -1,3 +1,5 @@
+# coding=utf-8
+
 class Temperature(object):
     '''Temperature facade.
 
--- a/plyer/platforms/linux/audio.py
+++ b/plyer/platforms/linux/audio.py
@@ -2,7 +2,7 @@ from plyer.facades.audio import Audio
 import subprocess
 
 
-class LinuxAudio(Audio):
+class GstreamerAudio(Audio):
     def __init__(self, file_path=None):
         default_path = 'test.mp3'
         super(LinuxAudio, self).__init__(file_path or default_path)
@@ -24,4 +24,4 @@ class LinuxAudio(Audio):
 
 
 def instance():
-    return LinuxAudio()
+    return GstreamerAudio()
--- a/plyer/platforms/linux/audio.py
+++ b/plyer/platforms/linux/audio.py
@@ -12,7 +12,6 @@ class LinuxAudio(Audio):
                                     "num-buffers=100", "!", "audioconvert",
                                     "!", "vorbisenc", "!", "oggmux", "!",
                                     "filesink", "location=" + self.file_path])
-        self.process.wait()
 
     def _stop(self):
         self.process.terminate()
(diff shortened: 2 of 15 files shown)


plyer/platforms/linux/audio.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, file_path=None):
+        default_path = '/home/recording.wav'
+        super(LinuxAudio, self).__init__(file_path or default_path)
+
+    def _start(self):
+        process = subprocess.Popen(("gst-launch", "autoaudiosrc",
+                                    "num-buffers=100", "!", "audioconvert",
+                                    "!", "vorbisenc", "!", "oggmux", "!",
+                                    "filesink", "location=" + self.file_path),
+                                   stdout=subprocess.PIPE)
```
malverick: add an extra parameter 'shell=True' to the `Popen` call.

plyer/platforms/linux/audio.py
```diff
@@ -0,0 +1,28 @@
...
+import subprocess
+
+
+class LinuxAudio(Audio):
+    def __init__(self, file_path=None):
+        default_path = '/home/recording.wav'
+        super(LinuxAudio, self).__init__(file_path or default_path)
+
+    def _start(self):
+        process = subprocess.Popen(("gst-launch", "autoaudiosrc",
```
malverick: should be `self.process`

plyer/platforms/linux/audio.py
```diff
@@ -0,0 +1,28 @@
...
+                                    "num-buffers=100", "!", "audioconvert",
+                                    "!", "vorbisenc", "!", "oggmux", "!",
+                                    "filesink", "location=" + self.file_path),
+                                   stdout=subprocess.PIPE)
+        process.wait()
```
malverick: `self.process.wait()`

plyer/platforms/linux/audio.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def _stop(self):
+        process.terminate()
```
malverick: `self.process.terminate()`

malverick:
@bhaveshAn initially when I compiled your code, logs generated were:
[initial logs](https://gist.github.com/malverick/6bcc14c63b5bb1b11c94e8098c955c4a)
After doing the above changes, logs are - 
[logs](https://gist.github.com/malverick/c339836c7a18ac8238d758bea978c1f6)

As I press the `Start Recording` button, I get `autoaudiosrc: 1: autoaudiosrc: gst-launch: not found` on terminal.
After I press the `Press to Stop Recording` button, I get ` OSError: [Errno 3] No such process` on terminal.

bhaveshAn:
@malverick Please try it using command line as well which is provided in facade.

bhaveshAn:
Tested and working fine on all versions of Linux Ubuntu.
However, this has not been merged so I am providing the link to [illustration of the feature](https://www.youtube.com/watch?v=dp4T5t-9mS8)
It does not having any dependency either. Hence in a merge-able state

plyer/facades/temperature.py
```diff
@@ -1,5 +1,6 @@
 # coding=utf-8
 
+
```
KeyWeeUsr: Same as previous.

plyer/platforms/linux/audio.py
```diff
@@ -0,0 +1,27 @@
+from plyer.facades.audio import Audio
+import subprocess
+
+
+class LinuxAudio(Audio):
```
KeyWeeUsr: This is definitely a GStreamer used as a provider for audio, not a "Linux" audio, so it'd nice to name it appropriately. There are other providers and I believe gnome or [kde](https://en.wikipedia.org/wiki/Amarok_(software)) has even its own minimalistic one.

akshayaurora:
this makes streamer a dependency...

Issue #312, opened 2017-03-18
No encoding and °C causes SyntaxError


zoltan-fedor:
Hi,
It seems that the docstring in `plyer/facades/temperature.py` was extended with the `°C` sign, but there is no encoding on the file, so it fails compile on Python 2 with the following error:
```
Obtaining plyer from git+https://github.com/kivy/plyer.git#egg=plyer
  Cloning https://github.com/kivy/plyer.git to ./venv1/src/plyer
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/Users/user/App/my-app/venv1/src/plyer/setup.py", line 4, in <module>
        import plyer
      File "plyer/__init__.py", line 16, in <module>
        from plyer import facades
      File "plyer/facades/__init__.py", line 38, in <module>
        from plyer.facades.temperature import Temperature
      File "plyer/facades/temperature.py", line 5
    SyntaxError: Non-ASCII character '\xc2' in file plyer/facades/temperature.py on line 6, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

Would it be possible to set encoding on the file when non-ascii characters are used?
Such as

```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
```

Thanks

cidermole:
Was merged in commit https://github.com/kivy/plyer/commit/c77c72e84fbe5aad8c6c9d24be9a8117c8fdb01c

tito:
Fixed in master

PR #313, opened 2017-03-18
Bugfix: set encoding for Python 2 compile

zoltan-fedor:
This file now has some UTF8 encoded docstrings, so it needs to set the charset.
See https://github.com/kivy/plyer/issues/312


zoltan-fedor:
Was fixed in the meantime by https://github.com/kivy/plyer/commit/74a0cfacf60e1ef427ee93f7d81155d4456daa5b

PR #315, opened 2017-03-22
OSX Audio feature added

bhaveshAn:
Audio Recording are done using videosnap and sox while audio being played by using afplay, a built in command line tool.


bhaveshAn:
I am looking to remove dependencies and if achieved then will make commits soon.

PR #316, opened 2017-03-22
[WIP] OSX Camera feature added

bhaveshAn:
Depends on ImageSnap and VideoSnap, if both are not available then Wacaw will be used.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,7 @@ Platform                           Android iOS Windows OS X Linux
 Accelerometer                      X       X           X    X
 Barometer                          X
 Call                               X       X
-Camera (taking picture)            X       X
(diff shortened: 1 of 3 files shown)


bhaveshAn:
I am looking to remove dependencies and if achieved then will make commits soon.

KeyWeeUsr:
If on GNU/Linux, it could seem to install random binaries might be a bearable approach. Hardly on macOS and definitely not with instructions like this:

> Copy the imagesnap file to someplace on your path like /usr/local/bin [...]

[`imagesnap`](https://github.com/rharder/imagesnap) and [`videosnap`](https://github.com/matthutchinson/videosnap) use a clear macOS API. Wacaw is GPL 2.0

PR #319, opened 2017-03-30
Linux ScreenShot feature added

bhaveshAn:
Not having dependencies either.
In a merge-able state.

screenshot


bhaveshAn:
screenshot

bhaveshAn:
Closed this pull request accidently. Sorry about that. Please check its running fine.
Tested on Linux Ubuntu 16.04 LTS.

kiok46:
@bhaveshAn thanks for contributing, please add `screenshotdemo.kv` file or load kivy string using `Builder.load_string()` in the example.
I have a suggestion, Why don't you create a new PR for screenshot for OSX, Linux and Windows? that way we could cover screenshots for all supported desktop devices and you won't have to recreate example in each PR (which you shouldn't). Also, you only need to add Windows as you already have done Linux and OSX.

bhaveshAn:
@kiok46 thanks for your review. Actually I am working on ScreenShot feature for Windows too but don't know how long it would take. So I can say that I have made the PR working from example too. If you merge this PR It will not required for me to recreate the example for screenshot in PR #324 and hence I will pull the upstream changes after deleting the example in PR of OSX. In this way, Workflow will remain same without a extra PR as you suggested.
Meanwhile providing proof of the working of feature from example too.

screenshot from 2017-04-07 00-10-06

screenshot from 2017-04-07 00-09-58

plyer/facades/screenshot.py
```diff
@@ -0,0 +1,62 @@
+'''
+ScreenShot
```
KeyWeeUsr: According to [Oxford dictionary](https://en.oxforddictionaries.com/definition/screenshot) a "screenshot" is a real word, unlike "FileChooser", therefore it'd be better to go for that. Also, I'd say it's more expected to do:

    from something import Screenshot

instead of:

    from something import ScreenShot

as it will be more confusing in the future. Also, it's inconsistent with the rest of your code, where you use "screenshot" as a single word.

plyer/facades/screenshot.py
```diff
@@ -0,0 +1,62 @@
...
+    '''
+    ScreenShot facade.
+    '''
+    _file_path = ''
+
+    def __init__(self, file_path):
+        super(ScreenShot, self).__init__()
+        self._file_path = file_path
+
+    def take_shot(self):
```
KeyWeeUsr: `screenshot.take()` seems visually more pleasant than `screenshot.take_shot()`

plyer/facades/temperature.py
```diff
@@ -1,5 +1,6 @@
 # coding=utf-8
 
+
```
KeyWeeUsr: Same as pr 320.

plyer/facades/screenshot.py
```diff
@@ -0,0 +1,62 @@
...
+    @property
+    def file_path(self):
+        return self._file_path
+
+    @file_path.setter
+    def file_path(self, location):
+        '''
+        Location of the screenshot.
+        '''
+        assert isinstance(location, (basestring, unicode)), \
```
KeyWeeUsr: Haven't tried it yet, but this seems smelly to me. Have you tried it both with py2 and py3?

plyer/platforms/linux/screenshot.py
```diff
@@ -0,0 +1,29 @@
+import subprocess
+from plyer.facades import ScreenShot
+from plyer.utils import whereis_exe
+
+
+class GnomeScreenShot(ScreenShot):
```
KeyWeeUsr: Perhaps a reference to the options (gnome, import) might be useful in the documentation?

bhaveshAn: It is done to give to use the feature with which-ever utitlity is present. So need is to give separate names to both implemented classes. What say @KeyWeeUsr

plyer/facades/screenshot.py
bhaveshAn: It is also have been used in the audio facade from long.

KeyWeeUsr: Definitely doesn't work on py3.

bhaveshAn: What say If I remove it. @KeyWeeUsr

akshayaurora:
This pr specifically uses gnome for screenshot. Assuming a Userinterface is something we can not do.

The way you check for ability to import Gnome then switch to using Gnome is something that can be used to test different interfaces like KDE or unity or ....

Please look into using XWindows directly to get the screenshot.

bhaveshAn:
@akshayaurora I have used command-line tools in my PRs because previously the features like Notification for linux is using tool like notify-send and Wifi is making use of nmcli tool which are pre-installed one like gst-launch-1.0 , gnome-screenshot, import , xrandr and bluetoothctl.

So, from my point of view there is no issues to use pre-installed binaries or command line tools to implement the feature, the idea that I acquire about plyer is to not having dependencies either.

If I am wrong please correct me. :)

akshayaurora:
Tools that are "universal & expected to be pre existing on most linux" distros are accepted if there is no obvious way around using them. In this case there is. XWindows & Xrandr extension both offer such options.

bhaveshAn:
@akshayaurora I browsed about it. find one such and running fine.
`sleep 5; xwd -root | convert - capture.png`

can i use it.

bhaveshAn:
@akshayaurora I have added the screenshot using xwd or XWindows in this branch at origin.
Can I make the pull request with the same branch or a new branch.
Please tell.

bhaveshAn:
Opened a new pull concerning linux screenshot at #326

PR #322, opened 2017-04-01
Pep8 fix in temperature.py

malverick:
fixes pep8 error in temperature facade. Should be merged because the error prevents plyer to install in system.


KeyWeeUsr:
Thanks!

PR #325, opened 2017-04-03
Linux Bluetooth feature

bhaveshAn:
Linux Bluetooth feature using `bluetoothctl` .  
methods in facade:
`bluetooth.enable()` to enable the bluetooth.
`bluetooth.disable()` to disable the bluetooth.
`bluetooth.visible()` to turn on the visibility of bluetooth.
`bluetooth.scan()` to start scanning the bluetooth devices.
`bluetooth.pair(name)` to pair the device with name=name.
`bluetooth.get_scan_devices()` to get the list of scanned devices.
`bluetooth.get_paired_devices()` to get the list of paired devices.
`bluetooth.connect(name)` to connect to the device with name=name.
`bluetooth.disconnect(name)` to disconnect with the device with name=name.

Proof of Implementation:

`bluetooth.enable()`   before enabling

enable_before

`bluetooth.enable() `  after enabling

enable_after

`bluetooth.visible()`   after turning on

visible

`bluetooth.get_scan_devices() `

get_scan_devices

`bluetooth.get_paired_devices()` 

get_paired_devices

`bluetooth.connect(name)`

connect

`bluetooth.disconnect(name)`

disconnect


mondlos:
Works great but even if your Bluetooth is already enabled you have to execute`bluetooth.enable()` before using the other commands, otherwise you will get the following error: `AttributeError: 'LinuxBluetooth' object has no attribute 'obj'`

bhaveshAn:
@mondlos 
>Works great

Thanks

> `bluetooth.send(self, msg)` would be a very useful method.

Tell the real idea of yours behind this. Since a standard feature of a common bluetooth is to send media to nearby device, only.

>And finally your code is very similar to this Gist: [https://gist.github.com/egorf/66d88056a9d703928f93](https://gist.github.com/egorf/66d88056a9d703928f93). I would give him credit and clarify if the licenses are compatible.

Off course the credit goes to [egorf](https:///github.com/egorf)
Since it is a gist, it has been used as same. Otherwise, I would have preserved his commits too.

KeyWeeUsr:
This PR is a code copied from this(https://gist.github.com/egorf/66d88056a9d703928f93) gist that clearly states the owner and a license of the code. The gist also states the code is a part of [ReachView](https://github.com/emlid/ReachView) project.

You just pasted such a code after **removing the license and the owner** - that is stealing. You clearly admitted you've taken the code from that particular gist, however you've taken it in a way that violates the very license [it was released under](https://github.com/emlid/ReachView/blob/master/LICENSE.txt). [1](https://github.com/emlid/ReachView/blob/master/LICENSE.txt#L46), [2](https://github.com/emlid/ReachView/blob/master/LICENSE.txt#L103), [3](https://github.com/emlid/ReachView/blob/master/LICENSE.txt#L175), [4](https://github.com/emlid/ReachView/blob/master/LICENSE.txt#L208) and other parts.

At the same time while MIT can be included in GPL projects most likely without any consequences (IANAL, it's [GPL compatible](https://www.gnu.org/licenses/license-list.html#Expat)), it doesn't work the other way around.

PR #330, opened 2017-04-13
Android Network Information

malverick:
Network Information API for android platform which describes the status of a network interface. 
Tested and working fine.
Tested on-
Device: Xiaomi Mi 4i
Android sdk: 20
Android api: 19
Android ndk: r9c
img-20170414-wa0001
img-20170414-wa0000

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,6 +44,7 @@ Audio recording                    X
 Flash                              X       X
 Wifi                                           X       X    X
 Temperature                        X
+Network Information                X
 ================================== ======= === ======= ==== =====
 
 Support
--- a/examples/networkinfo/buildozer.spec
+++ b/examples/networkinfo/buildozer.spec
@@ -1,0 +1,233 @@
+[app]
+
+# (str) Title of your application
+title = Network Information
+
+# (str) Package name
+package.name = plyernetworkinfo
+
+# (str) Package domain (needed for android/ios packaging)
+package.domain = org.test
+
+# (str) Source code where the main.py live
+source.dir = .
+
+# (list) Source files to include (let empty to include all the files)
+source.include_exts = py,png,jpg,kv,atlas
+
+# (list) List of inclusions using pattern matching
+#source.include_patterns = assets/*,images/*.png
+
+# (list) Source files to exclude (let empty to not exclude anything)
+#source.exclude_exts = spec
+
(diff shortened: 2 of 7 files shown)


KeyWeeUsr:
Maybe I'm just weird, but I feel like "network" might be easily confused with "wifi" or any other network, which in plyer's environment could mean confusing it even with LAN or basically confusing it with an option to control the whole networking on PCs:



Maybe a "cellular" would be a better alternative? @dessant what do you think? If it's just too "off", then ignore this suggestion 😋

dessant:
I'd go as far as not be sure about the legitimacy of this as a plyer facade. 😝

kiok46:
IMO, this could be nice add for plyer.
@KeyWeeUsr since [get_type_name](https://github.com/kivy/plyer/pull/330/files#diff-9f21429e259074cccbd9ecd6f4102088R27) method shows the type of network, WIFI, ETHERNET etc. then "Connectivity" could also be an option, what do you say? :) 
I also want you guys to see:  [Types of network in Android Connectivity Manager](https://developer.android.com/reference/android/net/ConnectivityManager.html#TYPE_BLUETOOTH)

KeyWeeUsr:
+1 for "connectivity" too. Reading about what does it have, this facade might even return booleans for available networks e.g. mobile, wifi, bluetooth, etc. On the other hand, we already have [such thing for wifi](https://github.com/kivy/plyer/blob/master/plyer/facades/wifi.py#L69), similar for IRblaster and if we'd include other facades e.g. bluetooth, then this facade _might_ become totally useless.

Which gets me to an argument for the "cellular" - if this facade would even support setting some values such as airplane mode, switching SIMs or something in that sense, then it'd probably have more use-cases. Also, iirc there was an option for a PC to communicate with phones via PuTTy to send/recieve SMS, make calls, etc like ~5-7 years ago. Nowadays probably no one sane uses that, but it _could_ mean there's even something for desktops. Probably.

bhaveshAn:
I think the network information should state the IP address, SUBNET mask, GATEWAY, DNS Information also as without these information, to have a network is not preferable.

malverick:
If `cellular` is the main focus here then I can use [Telephony Manager](https://developer.android.com/reference/android/telephony/TelephonyManager.html) class which contains all the methods related to `network` and `SIM`. @KeyWeeUsr what say?

PR #331, opened 2017-04-14
Windows Screenshot feature

bhaveshAn:
Tested and working fine. Kindly merge this. 

screenshot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ venv
 bin
 .buildozer
 .idea
+*.jpg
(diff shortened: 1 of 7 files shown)


KeyWeeUsr:
This PR uses `PIL` or `Pillow` i.e. a dependency. On the other hand, Windows provides an API extensively explained [here](https://docs.microsoft.com/en-us/windows/desktop/gdi/capturing-an-image). We don't need the whole thing, just a simple desktop Window screen capture. I'll take a look into implementing that API.

PR #335, opened 2017-05-13
Update win_api_defs.py

malverick:
No use of having `ctypes.` in  `ctypes.Structure` or `ctypes.POINTER` or `ctypes.windll` when the corresponding libraries are already imported above.


akshayaurora:
@malverick thanks

malverick:
Your welcome @akshayaurora :)

PR #337, opened 2017-05-14
Gyroscope uncalibrated sensor

malverick:
Methods for accessing gyroscope uncalibrated sensor are added in current gyroscope API.
Example for gyroscope api is also created.

Tested and working fine. Can be merged.

### DIFF
--- a/examples/gyroscope/buildozer.spec
+++ b/examples/gyroscope/buildozer.spec
@@ -1,0 +1,233 @@
+[app]
+
+# (str) Title of your application
+title = Gyroscope
+
+# (str) Package name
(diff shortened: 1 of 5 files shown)


malverick:
Added the iOS implementation to retrieve uncalibrated values.

malverick:
I have removed the mentioning the unit of readings received from the sensor because in `Android` the readings are in `rad/s` but in iOS there is no mentioning of the unit of readings so it is best to not mention it at all.

akshayaurora:
@malverick thanks 👍

malverick:
@akshayaurora your welcome :)

PR #343, opened 2017-06-17
Fix bug in WindowsBalloonTip

Chronial:
`icon_path_name` is not defined. So passing an invalid icon file will raise an error about that instead of the expected Exception:
```
  File ".\win-popup.py", line 6, in <module>
    app_name='sync-seedbox', timeout=10)
  File "C:\Tools\Python35\lib\site-packages\plyer\platforms\win\libs\balloontip.py", line 145, in balloon_tip
    WindowsBalloonTip(**kwargs)
  File "C:\Tools\Python35\lib\site-packages\plyer\platforms\win\libs\balloontip.py", line 91, in __init__
    format(icon_path_name))
NameError: name 'icon_path_name' is not defined
```

### DIFF
--- a/plyer/platforms/win/libs/balloontip.py
+++ b/plyer/platforms/win/libs/balloontip.py
@@ -90,8 +90,7 @@ class WindowsBalloonTip(object):
             hicon = win_api_defs.LoadImageW(None, app_icon, IMAGE_ICON, 0, 0,
                                             icon_flags)
             if hicon is None:
-                raise Exception('Could not load icon {}'.
-                                format(icon_path_name))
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
@Chronial Thanks! It's basically from https://github.com/kivy/plyer/commit/1d7fb1fcc1d0171de35aede4060a54e3f8df5792 where the name was removed, but apparently not updated(typo). The `else` part of that block should load just the default application icon:

> IDI_APPLICATION MAKEINTRESOURCE(32512) | Default application icon.
([source](https://msdn.microsoft.com/en-us/library/windows/desktop/ms648072(v=vs.85).aspx))

PR #362, opened 2017-07-28
iOS Proximity Sensor

malverick:


plyer/platforms/ios/proximity.py
```diff
@@ -0,0 +1,37 @@
...
+        self.device.setProximityMonitoringEnabled_(False)
+
+    def _get_proximity(self):
+        if self.device.proximityMonitoringEnabled:
+            if self.device.proximityState:
+                return str(True)
+            else:
+                return str(False)
+        else:
+            return 'Proximity Sensor in present in your device.'
```
tshirtman: "is not present" ?

plyer/platforms/ios/proximity.py
```diff
@@ -0,0 +1,37 @@
...
+    def _enable(self):
+        self.device.setProximityMonitoringEnabled_(True)
+
+    def _disable(self):
+        self.device.setProximityMonitoringEnabled_(False)
+
+    def _get_proximity(self):
+        if self.device.proximityMonitoringEnabled:
+            if self.device.proximityState:
+                return str(True)
```
tshirtman: Why now returning a string of a boolean, both platforms are coded to return a boolean value anyway?

examples/proximity/main.py
```diff
@@ -37,7 +37,7 @@
     Label:
         text: 'Does Proximity Sensor detect something?'
     Label:
-        text: 'Yes' if root.is_near else 'No'
+        text: root.is_near
```
malverick: @tshirtman I was trying to remove this condition statement from here. That's why String Property.

tshirtman: I don't think it's a good idea to change the api to simplify a particular example, especially since a lot of other usages would change from:

    if obj.is_near:
        do_something()

to

    if obj.is_near == 'True':
         do_something()

which is really un-pythonic, the api is better when a boolean value is represented as a boolean.

tshirtman: Also, you could just change the example to this, to achieve the same effect as your previous solution.

    text: str(root.is_near)

akshayaurora: "text: str(root.is_near)"
👍

PR #368, opened 2017-09-10
Warn instead stder.write for linux notif errors

sametmax:
It does the same thing, except you can silence it if you wish. Given that you may very well not want to ever use the db-bus binding, it's nice to have the choice of sticking to notify-send without any mention of it in stder if you feel like it.


KeyWeeUsr:
Thanks! :)

PR #370, opened 2017-09-20
Update buildozer.spec

sandeepsajan0:
now camera facade of plyer(kivy) works well on android also.

### DIFF
--- a/examples/camera/basic/buildozer.spec
+++ b/examples/camera/basic/buildozer.spec
@@ -31,7 +31,9 @@ source.include_exts = py,png,jpg,kv,atlas
 # (str) Application versioning (method 2)
 version = 1.0
 
-# (list) Application requirements
+# (list) Application requirements 
+# android library is also required to run this app on Android platform
+# for android device -> requirements = plyer,kivy,android
 requirements = plyer,kivy
 
 # (str) Presplash of the application
--- a/examples/accelerometer/basic/buildozer.spec
+++ b/examples/accelerometer/basic/buildozer.spec
@@ -32,6 +32,7 @@ source.include_exts = py,png,jpg,kv,atlas
 version = 1.0
 
 # (list) Application requirements
+#for android device-> requirements=plyer,kivy,android
 requirements = plyer,kivy
 
 # (str) Presplash of the application


sandeepsajan0:
sorry, I do not understand what's going here please anyone tell me

kiok46:
@sandeepsajan0  It's just a bot. To know more visit [coverals.io](https://coveralls.io/).

examples/camera/basic/buildozer.spec
```diff
@@ -31,8 +31,9 @@ source.include_exts = py,png,jpg,kv,atlas
 # (str) Application versioning (method 2)
 version = 1.0
 
-# (list) Application requirements
-requirements = plyer,kivy
+# (list) Application requirements 
+# android is also require for run this app on android
```
malverick: Grammar mistake :P . `android library is also required to run this app on Android platform` might be better.

examples/camera/basic/buildozer.spec
```diff
@@ -31,8 +31,9 @@ source.include_exts = py,png,jpg,kv,atlas
...
+requirements = plyer,kivy,android
```
malverick: I don't think there is need to mention android in `requirements` since this buildozer.spec can also be used to create iOS app. Just informing the user with the help of comment will be enough. 
`requirements` should contain the common libraries which are required for both Android and iOS.

sandeepsajan0:
@malverick  I think it will be fine now.

examples/camera/basic/buildozer.spec
sandeepsajan0: ok, I have done it.
now it should be fine.

sandeepsajan0:
Now, I think it should be merge, if not tell me the reason, please.

akshayaurora:
I understand what you are trying to do here, but right now this is just a comment. How about a new pr? where you actually add `android` to the requirements line.

Start with a new branch and do a new pr with just one commit, plus add comments on why this change was needed.

looking forward to your new pr.

sandeepsajan0:
Yes, you are right but if I actually add  `android` in requirements then it will not work for IOS and if I add a comment for IOS then it will be the same as above because in it I had added a comment for android.

akshayaurora:
ok, then please squash the commits into one. This would help us get a clean history.

KeyWeeUsr:
@akshayaurora You can still change the merging type to `Squash and merge` and that will add only a single commit to our repo. I'm not sure if the comment in .spec file would help. Perhaps we should add it to some p4a recipe as a hard requirement or switch to pyjnius completely if possible.

sandeepsajan0:
@akshayaurora  can I squash the commits of closed pull request?

KeyWeeUsr:
@sandeepsajan0 afaik yeah, but if you push them back to remote (`git push -f`), the PR might lock (happened to me long ago). Reopening.

KeyWeeUsr:
Thanks! :)

Issue #369, opened 2017-09-20
Using the camera crashes the app


sstoychev:
I have kivy app and using plyer camera:
```
...
from plyer import camera
...
    def open_camera(self):
        self.dismiss_popup()

        # Take a picture and save at this location. After will call done() callback
        pics_dir = os.path.join(App.get_running_app().user_data_dir, 'pics')
        if not os.path.exists(pics_dir):
            os.makedirs(pics_dir)
        pic_name = os.path.join(pics_dir, 'main.jpg')
        camera.take_picture(pic_name, self.done)
        self.additional_info.text = pic_name

    def done(self, e):
        print(e)
...

```
When I click to open the camera the app crashes.
```
...
09-20 02:34:19.031  7285  7410 W PythonActivity: Accessing org.renpy.android.PythonActivity is deprecated and will be removed in a future version. Please switch to org.kivy.android.PythonActivity.
09-20 02:34:19.061  7285  7410 F art     : art/runtime/java_vm_ext.cc:410] JNI DETECTED ERROR IN APPLICATION: static jfieldID 0xb05b16c0 not valid for class java.lang.Class<org.renpy.android.PythonActivity>
09-20 02:34:19.061  7285  7410 F art     : art/runtime/java_vm_ext.cc:410]     in call to GetStaticObjectField
09-20 02:34:19.061  7285  7410 F art     : art/runtime/java_vm_ext.cc:410]     from int org.libsdl.app.SDLActivity.nativeInit(java.lang.Object)
09-20 02:34:19.061  7285  7410 F art     : art/runtime/java_vm_ext.cc:410] "SDLThread" prio=5 tid=12 Runnable
09-20 02:34:19.061  7285  7410 F art     : art/runtime/java_vm_ext.cc:410]   | group="main" sCount=0 dsCount=0 obj=0x12c514c0 self=0x9ed98400
09-20 02:34:19.061  7285  7410 F art     : art/runtime/java_vm_ext.cc:410]   | sysTid=7410 nice=0 cgrp=default sched=0/0 handle=0x9dd70930
09-20 02:34:19.061  7285  7410 F art     : art/runtime/java_vm_ext.cc:410]   | state=R schedstat=( 5939062751 1246171278 5527 ) utm=533 stm=60 core=0 HZ=100
09-20 02:34:19.061  7285  7410 F art     : art/runtime/java_vm_ext.cc:410]   | stack=0x9dc6e000-0x9dc70000 stackSize=1038KB
[...]

Issue #338, opened 2017-05-30
Example Applications break down on a real device


burkouski:
Hi guys. I tried to deploy some of examples on a real device (android 6) using buildozer. At first I tried camera example, but app breaks down when I tried take a picture with following error:

`DEBUG   : Abort message: 'art/runtime/java_vm_ext.cc:410] JNI DETECTED ERROR IN APPLICATION: static jfieldID 0xb052a4b0 not valid for class java.lang.Class<org.renpy.android.PythonActivity>'
DEBUG   :  r0 00000000  r1 00000c03  r2 00000006  r3 a0fa9978
DEBUG   :     r4 a0fa9980  r5 a0fa9930  r6 0000000b  r7 0000010c
DEBUG   :     r8 b4b3f378  r9 b4d2a800  sl 00000001  fp 00000001
DEBUG   :     ip 00000006  sp a0fa3200  lr b6d16c41  pc b6d19030  cpsr 40070010
DEBUG   : 
DEBUG   : backtrace:
DEBUG   :     #00 pc 00042030  /system/lib/libc.so (tgkill+12)
DEBUG   :     #01 pc 0003fc3d  /system/lib/libc.so (pthread_kill+32)
DEBUG   :     #02 pc 0001c3db  /system/lib/libc.so (raise+10)
DEBUG   :     #03 pc 0001958d  /system/lib/libc.so (__libc_android_abort+34)
DEBUG   :     #04 pc 000174e0  /system/lib/libc.so (abort+4)
DEBUG   :     #05 pc 00333a39  /system/lib/libart.so (_ZN3art7Runtime5AbortEv+228)
DEBUG   :     #06 pc 000f464b  /system/lib/libart.so (_ZN3art10LogMessageD2Ev+2226)
DEBUG   :     #07 pc 0025aa9f  /system/lib/libart.so (_ZN3art9JavaVMExt8JniAbortEPKcS2_+1550)
DEBUG   :     #08 pc 0025ae4d  /system/lib/libart.so (_ZN3art9JavaVMExt9JniAbortVEPKcS2_St9__va_list+64)
DEBUG   :     #09 pc 000fd221  /system/lib/libart.so (_ZN3art11ScopedCheck6AbortFEPKcz+32)
[...]

PR #373, opened 2017-10-09
update readme

sandeepsajan0:
It will be helpful to run the facades examples on android also. 
Any user can change the .spec file after read the README file.
It also can fix some issues like #369 #338(partially).


akshayaurora:
@sandeepsajan0 thanks

PR #377, opened 2017-11-02
Arrange APIs table in alphabetical order

malverick:


malverick:
@kiok46 is this fine?

PR #388, opened 2018-03-12
enhancement: bluetooth status

kapilnayar:
the code changes and the example has been tested in os-x.


akshayaurora:
@kapilnayar thanks a lot  👍

PR #390, opened 2018-04-13
Wifi upgrade

kuzeyron:
This have been my main focus.
Less Sudo and packages!

### DIFF
--- a/plyer/platforms/linux/wifi.py
+++ b/plyer/platforms/linux/wifi.py
@@ -13,8 +13,6 @@ import gi
 gi.require_version('NM', '1.0')
 from gi.repository import NM
 
-nmc = NM.Client.new()
-devs = nmc.get_devices()
 def ssid_to_utf8(ap):
 	ssid = ap.get_ssid()
 	if not ssid:
@@ -97,6 +95,8 @@ class LinuxWifi(Wifi):
 			return True
 		return False
 	def _start_scanning(self):
+		nmc = NM.Client.new()
+		devs = nmc.get_devices()
 		i = 0
 		hosts = {}
 		for dev in devs:


examples/wifi/main.py
```diff
@@ -106,8 +106,12 @@ def show_wifi_scans(self):
         for name in wifi_scans:
             content = ""
             items = wifi._get_network_info(name)
-            for key, value in items.iteritems():
-                content += "{}:    {} \n".format(key, value)
+            if sys.version_info.major is 3: #Support for Py2 and Py3
+                for key, value in items.items():
+                    content += "{}:    {} \n".format(key, value)
+            else:
+                for key, value in items.iteritems():
```
inclement: I would just use `items.items()` to keep things simple, it seems unlikely that the dict will have enough keys for it to matter in Python 2.

plyer/platforms/linux/wifi.py
```diff
@@ -4,37 +4,76 @@
         - nmcli (Network Manager command line tool)
             It is found in most of the popular distros. Support for other
             managers is not provided yet.
-
-        - python-wifi module
-            `https://wifi.readthedocs.io/en/latest/`
-            `https://github.com/rockymeza/wifi`
-
+    One more thing: Betalistish!!
```
inclement: Did you mean to include this?

plyer/platforms/linux/wifi.py
```diff
@@ -73,26 +112,24 @@ def _connect(self, network, parameters):
...
     import sys
     try:
-        import wifi
         return LinuxWifi()
     except ImportError:
-        sys.stderr.write("python-wifi not installed. try:"
-                         "`pip install --user wifi`.")
-
-    return Wifi()
+        sys.stderr.write("Oh look a penny..")
```
inclement: Did you mean to include this?

examples/wifi/main.py
kuzeyron: I think I tried using only items.items(). But if someone have time to clean my mess I'd be more than happy!

plyer/platforms/linux/wifi.py
kuzeyron: Only for beta testers! There are some jobs left to do. But beta testers are needed right now!

kuzeyron:
Changed the code to work better.
Source taken from: https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/examples/python/gi/show-wifi-networks.py

plyer/platforms/linux/wifi.py
```diff
@@ -1,98 +1,186 @@
...
+		- nmcli (Network Manager command line tool)
+			It is found in most of the popular distros. Support for other
+			managers is not provided yet.
+			Source from: https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/examples/python/gi/show-wifi-networks.py
 '''
-
+#from __future__ import print_function
 from plyer.facades import Wifi
 from subprocess import Popen, PIPE, call, STDOUT
+import gi
```
akshayaurora: This is a strict no, we can not just import a gui lib for managing wifi `nmcli` is not dependent on the gui lib. Think of distress that have no gnome libs at all.

akshayaurora:
closing due to dependency on go

PR #396, opened 2018-04-28
why should decode 'l' again and again

xhimanshuz:
If output object can be decoded once.


plyer/platforms/linux/battery.py
```diff
@@ -18,19 +17,19 @@ def _get_state(self):
             ["upower", "-d", dev],
             stdout=PIPE
         )
-        output = upower_process.communicate()[0]
-
+        output = upower_process.communicate()[0].decode()
+
```
akshayaurora: please remove the extra space.

akshayaurora:
@xhimanshuz thanks

PR #405, opened 2018-07-01
Fix all pep8 errors in the repository

Sires0:
Only more **a lot of errors but still fits my screen** to go

Ok, fixed the mixed commits issue

I'm finishing to fix all PEP8, then I'm gonna fix the errors caused...


plyer/platforms/linux/keystore.py
```diff
@@ -1,6 +1,6 @@
 try:
     import keyring
-except:
+except Exception:
```
tshirtman: Ok, not sure if we can do better there, but that would be nice, excepting on `Exception` is equivalent to just using a bare `except`, so it's a bit bad to do that to follow a style guide.

plyer/platforms/macosx/libs/osx_paths.py
```diff
@@ -1,11 +1,15 @@
 import ctypes
 import os
 
+LibraryPath = \
+    "/System/Library/Frameworks/CoreFoundation.framework/\
+    Versions/A/CoreFoundation"
```
tshirtman: i think this is a breaking change here, the spaces will be part of the string, you could use something like
```
LibraryPath = ''.join(
    "/System/Library/Frameworks/CoreFoundation.framework/",
   "Versions/A/CoreFoundation"
)
```

or even simply
```
LibraryPath = (
    "/System/Library/Frameworks/CoreFoundation.framework/"
    "Versions/A/CoreFoundation"
)
```
(note the lack of comma at the end of the line, so strings are simply concatenated)

Sires0:
Changes made, please remove your "requested changes", I'm near ready to merge rn.
I just don't understand the why of the assert error at appveyor python 3.6

plyer/platforms/macosx/libs/osx_paths.py
```diff
@@ -1,11 +1,14 @@
 import ctypes
 import os
 
+LibraryPath = "/System/Library/Frameworks/CoreFoundation.framework/"
+"Versions/A/CoreFoundation"
```
tshirtman: Does this work? i'm pretty sure the second line won't be bart of the string.

```
>>> a = "aaa"
>>> "bbb"
'bbb'
>>> print(a)
aaa
```
but you could just add a `\` at the end of the first line.
```
>>> a = "aaa"\
... "bbb"
>>> print(a)
aaabbb
```

Sires0: ok

Sires0:
I'd say it's ready to merge, appveyor assert errors seems to be internal, since one time it gives on PEP8, then on PY2 but not on PEP8, AppVeyor seems way 2 broken. And it's not even related to my code

tshirtman:
I rebased and merged manually, thanks!

PR #424, opened 2018-10-01
Switch from Ubuntu Trusty to Ubuntu Bionic LTS

KeyWeeUsr:
Changing to Bionic LTS as per @AndreMiras' suggestion.


AndreMiras:
Great! By the way I didn't know Travis had the `dist` `bionic`

Issue #282, opened 2017-02-09
Need Audio recording for iOS, OSX.


kiok46:
Need audio recording feature for both iOS and OSX.
One closed PR #216  might give some lead to work on it but some methods are not implemented well. Feel free to use that code.

Nephyx:
Going to try implement this for macOS.

PR #428, opened 2018-10-06
Audio recording for macOS

Nephyx:
Added audio recording and playback for macOS using Apple's Objective-C API via pyobjus.
Audio files are recorded with 44.1 kHz sample rate in WAV format.

Closes issue #282.


plyer/platforms/macosx/audio.py
```diff
@@ -0,0 +1,81 @@
...
+        # Audio recorder instance initialization with specified file NSURL
+        # and audio file format
+        self._recorder = AVAudioRecorder.alloc()
+        self._recorder = self._recorder.initWithURL_format_error_(
+            file_NSURL, af, None
+        )
+        
+        if self._recorder:
+            self._recorder.record()
+            print("Recording...")
```
tito: Please remove all the print in the implementation

plyer/platforms/macosx/audio.py
```diff
@@ -0,0 +1,81 @@
...
+
+AVAudioPlayer = autoclass("AVAudioPlayer")
+AVAudioRecorder = autoclass("AVAudioRecorder")
+AVAudioFormat = autoclass("AVAudioFormat")
+NSString = autoclass('NSString')
+NSURL = autoclass('NSURL')
+
+class OSXAudio(Audio):
+    def __init__(self, file_path=None):
+        default_path = join(expanduser('~'), 'Desktop', 'audio.wav')
```
tito: You should not pass a default value that point on nothing :) Should be None by default.

plyer/platforms/macosx/audio.py
```diff
@@ -0,0 +1,81 @@
...
+            self._player.stop()
+            print("Stopped playing.")
+            self._player = None
+
+    def _play(self):
+        # Audio player instance initialization with the file NSURL
+        # of the last recorded audio file
+        self._player = AVAudioPlayer.alloc()
+        self._player = self._player.initWithContentsOfURL_error_(
+            self._current_file, None
```
tito: It would be better to catch the error and raise an exception if something failed, rather than completely ignore.

plyer/platforms/macosx/audio.py
Nephyx: Tried without passing a value of `None`, but I started getting the following error:
`>>> from plyer import audio
>>> audio.start()
  File "/Users/maroskovac/Dev/plyer/plyer/utils.py", line 96, in _ensure_obj
    obj = mod.instance()
  File "/Users/maroskovac/Dev/plyer/plyer/platforms/macosx/audio.py", line 82, in instance
    return OSXAudio()
TypeError: __init__() takes exactly 2 arguments (1 given)
  File "/Users/maroskovac/Dev/plyer/plyer/facades/audio.py", line 65, in start
    self._start()
  File "/Users/maroskovac/Dev/plyer/plyer/facades/audio.py", line 98, in _start
    raise NotImplementedError()
NotImplementedError`

Nephyx:
@tito here's a quick summary of my commits addressing your changes request:
- removed all prints
- kept passing `file_path` with default value of `None`, as removing it as per your suggestion causes `NotImplementedError` (more in my reply to your change request)
- added raising exceptions in case of recorder or player initialization errors (using only one NSError instance because of segmentation faults in case of multiple NSError allocs and actually, there's no need to have more than one instance)

tito:
Thanks you!

PR #431, opened 2018-10-08
Changed OSX storage path from ctypes to pyobjus

Nephyx:
Changed OSX storage path implementation from using ctypes to pyobjus because of ctypes returning malformed path strings (probably null characters).
The implementation with the same Apple's Objective-C API as the one for iOS, OSX version works.

Related to issue #282 and pull request #428.


tito:
Looks good to me, thanks you!

PR #435, opened 2018-10-12
Added iOS audio recording implementation

Nephyx:
Added audio recording and playback for iOS using Apple's Objective-C API via pyobjus.
Audio files are recorded with 44.1 kHz sample rate in WAV format.

Related to pull request #428.
Closes issue #282.

### DIFF
--- a/plyer/platforms/ios/audio.py
+++ b/plyer/platforms/ios/audio.py
@@ -1,0 +1,85 @@
+from plyer.facades import Audio
+from pyobjus.dylib_manager import load_framework, INCLUDE
+from pyobjus import autoclass
+from os.path import join
+from plyer.platforms.ios.storagepath import iOSStoragePath
+
+load_framework(INCLUDE.Foundation)
+load_framework(INCLUDE.AVFoundation)
+
+AVAudioPlayer = autoclass("AVAudioPlayer")
+AVAudioRecorder = autoclass("AVAudioRecorder")
+AVAudioFormat = autoclass("AVAudioFormat")
+NSString = autoclass('NSString')
+NSURL = autoclass('NSURL')
+NSError = autoclass('NSError').alloc()
+
+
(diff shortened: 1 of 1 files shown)


orrburgel:
Thank you very much, I was looking for this, I was actually looking for STT solution that will work with iOS but then I figured that I might be able to do it myself by sending recordings to an external server, Is there any plans for an iOS VTT solution?
According to the Supported APIs table Speech to Text (STT) only works on Android but shouldn't it also work on Windows using something like this?: https://pypi.org/project/SpeechRecognition/1.2.3/

banasiakmichal:
Hi there, guys do You still support audio recording for iOS (?)
I checked out code from https://github.com/Nephyx/plyer/blob/ios_audio/plyer/platforms/ios/audio.py. 
Basically there is some issues ( iOS 15.2, iPhone 7). The output looks like this:


[INFO   ] [Base        ] Start application main loop
2022-02-08 15:12:41.751431+0200 domi[8191:4020460] Unbalanced calls to begin/end appearance transitions for <SDL_uikitviewcontroller: 0x7f7be77cdd80>.
2022-02-08 15:12:41.752343+0200 domi[8191:4020460] Unbalanced calls to begin/end appearance transitions for <SDL_uikitviewcontroller: 0x7f7be77cdd80>.
2022-02-08 15:13:20.228573+0200 domi[8191:4020460] [plugin] AddInstanceForFactory: No factory registered for id <CFUUID 0x6000038b3ca0> F8BB1C28-BAE8-11D6-9C31-00039315CD46
2022-02-08 15:13:20.286350+0200 domi[8191:4020460]  CreateDataFile failed
2022-02-08 15:13:20.287093+0200 domi[8191:4020460]  Couldn't create a new audio file object
(lldb) 

is it a problem with device memory ?? Could You give any tips ?

PR #437, opened 2018-10-13
Update wifi.py

### LINKED ISSUE #436: plyer.wifi.is_enabled() always return false running in python3
**[plyer/platforms/linux/wifi.py](https://github.com/kivy/plyer/blob/master/plyer/platforms/linux/wifi.py)** always returns **False** when it's _is_enabled()_ function is called.
<br>
Python3: Popen object's communicate() function return a tuple with it's first value like b'<str:enable_status>'
<br>
Problem:
[Line#26](https://github.com/kivy/plyer/blob/master/plyer/platforms/linux/wifi.py#L26) - The value is checked directly against the normal string.

Vibhu-Agarwal:
Fixes #436 
The **decode()** function fixes the problem for both Python2 and Python3


KeyWeeUsr:
Thank you! :dango:

PR #438, opened 2018-10-13
Add extra options for installation via setuptools

KeyWeeUsr:
More comfortable than using `try`/`except` and shut the user down with a warning. Instead we should provide an interface to install the conditional requirement, i.e. in this case `pyjnius` for Android platform and `pyobjus` for macOS/iOS.

Choosing non-intrusive installation option because it might break stuff for mobile platforms if enforcing the requirements on pyobjus/pyjnius + there are no releases of pyobjus/pyjnius yet (wip), therefore:

    pip install plyer[macosx]

will for now just crash. The `dev` option will work though.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,8 +15,21 @@ For Android install additional dependency PyJNIus::
 from os.path import dirname, join
 import plyer
 
+EXTRA_OPTIONS = {}
+
 try:
     from setuptools import setup
+    EXTRA_OPTIONS = dict(
+        EXTRA_OPTIONS, **{
+            'extras_require': {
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Ref https://github.com/kivy/plyer/commit/3892b41a8d9a8006754d95c4dfc2903f12d16bf0#r29926423, but we still need pyobjus released and for that we need `CArray` fixed ([tests failed](https://travis-ci.org/kivy/pyobjus/jobs/441421231)) at least and then port it to py3. I'll push my work for py3 to new branch in pyobjus and perhaps someone can figure out how to push it further.

PR #447, opened 2018-10-14
Cron deploy

KeyWeeUsr:
Add a build section to the `plyer.__version__` with standard .dev0 string. Versions are fetched from PyPI according to the current version of plyer. If there has been a release e.g. 1.2.3 and we bump the version on master to 1.2.4.dev0, this fetches all versions that start with 1.2.4 i.e. right after the release there are none and via cronjob there will be a 1.2.4.0.dev0 release. Afterwards 1.2.4.1.dev0, 1.2.4.2.dev0, etc until we bump to e.g. 1.2.5.dev0.

Event type according to https://docs.travis-ci.com/user/environment-variables/.

Current setup on Travis (does nothing serious until this is merged):

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -127,6 +127,7 @@ script:
                       --tty
                       --env COVERALLS_REPO_TOKEN=$COVERALLS_REPO_TOKEN
                       --env COVERALLS_SERVICE_NAME=travis-ci
+                      --env TRAVIS_EVENT_TYPE=$TRAVIS_EVENT_TYPE
                       --env TRAVIS_JOB_ID=$TRAVIS_JOB_ID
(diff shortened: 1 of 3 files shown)


KeyWeeUsr:
I wonder whether this even is a right approach to choose because we have multiple platforms for testing and we'd still end up with testing it before a stable release + through this we'd have an unnecessary clutter on PyPI which doesn't sound good. Also the current approach doesn't handle the same commit, therefore if the repo is left untouched for 2 months, there'd still be releases.

Closing, since I'm not happy with the idea of producing clutter on PyPI, but I'm still for doing the releases more often (e.g. monthly), but manually with all the testing, since now we only need `PLYER_DEPLOY=1` on Travis for a new release.

PR #465, opened 2018-11-05
fetching device interface name (ifname)

xhimanshuz:
Now device interface supports the new naming scheme of a network interface. Hardcode interface name not needed anymore.

### DIFF
--- a/plyer/platforms/linux/wifi.py
+++ b/plyer/platforms/linux/wifi.py
@@ -27,6 +27,10 @@ class LinuxWifi(Wifi):
             break
     else:
         ifname = "wlan0"
+        
+    com = Popen(["nmcli","--version"],stdout=PIPE)
+    com = com.communicate()[0].strip()
+    nmcliVersion = int(com.split(" ")[-1].replace(".", ""))
     
     def _is_enabled(self):
         '''
@@ -98,7 +102,10 @@ class LinuxWifi(Wifi):
         '''
         Disconnect all the networks managed by Network manager.
         '''
-        return call(['nmcli', 'nm', 'enable', 'false'])
+        if(self.nmcliVersion > 98):
+            return call(['nmcli','networking','off'])
+        else:
+            return call(['nmcli', 'nm', 'enable', 'false'])
 
 
 def instance():
--- a/plyer/platforms/linux/wifi.py
+++ b/plyer/platforms/linux/wifi.py
@@ -18,7 +18,8 @@ from subprocess import Popen, PIPE, call
 class LinuxWifi(Wifi):
     names = {}
     
-    ifname = Popen(["nmcli", "-t","connection", "show", "--active"], stdout=PIPE, stderr=PIPE).communicate()[0].decode().split(":")[-1].strip()
+    Popen(["nmcli", "-t","connection", "show", "--active"], stdout=PIPE, stderr=PIPE).
+    ifname = communicate()[0].decode('utf-8').split(":")[-1].strip()
     
     def _is_enabled(self):
         '''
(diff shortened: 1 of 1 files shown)


plyer/platforms/linux/wifi.py
```diff
@@ -17,7 +17,10 @@
 
 class LinuxWifi(Wifi):
     names = {}
-
+    
+    Popen(["nmcli", "-t","connection", "show", "--active"], stdout=PIPE, stderr=PIPE).
```
KeyWeeUsr: You have a dot (`.`) at the end :)

Anyway, please don't pipe `stderr`, otherwise it's the same like `try: <stuff> except: pass` i.e. just eating any error the binary might raise. It's better to actually have something in the log than nothing. :)

plyer/platforms/linux/wifi.py
```diff
@@ -17,7 +17,10 @@
...
+    ifname = communicate()[0].decode('utf-8').split(":")[-1].strip()
```
KeyWeeUsr: While this seems nice in general, `communicate()` is a method of an object returned from `Popen()` and you call it out of nowhere.

Run your code first before committing and also include a default `wlan0` value in case the `Popen()` call fails with non-zero error code and/or invalid value because the original implementation uses old arguments in the `call()`, therefore your way of fetching the interface might not be available in the old `nmcli` binary.

xhimanshuz: Sorry, i was on my tablet so, i commit directly. :p

plyer/platforms/linux/wifi.py
```diff
@@ -17,7 +17,21 @@
 
 class LinuxWifi(Wifi):
     names = {}
-
+    
+    com = Popen(["nmcli", "-t","connection", "show", "--active","--order","active" ], stdout=PIPE )
+    com = com.communicate()[0].strip().split("\n")
+    for i in com:
```
KeyWeeUsr: This isn't good approach. Logic such as this should be container within a function e.g. a constructor (`__init__` in python). Refactor it please.

plyer/platforms/linux/wifi.py
```diff
@@ -17,7 +17,21 @@
 
 class LinuxWifi(Wifi):
     names = {}
-
+    
+    com = Popen(["nmcli", "-t","connection", "show", "--active","--order","active" ], stdout=PIPE )
+    com = com.communicate()[0].strip().split("\n")
```
KeyWeeUsr: This will be a problem if you don't decode.

plyer/platforms/linux/wifi.py
```diff
@@ -17,7 +17,21 @@
...
+    for i in com:
+        if("wireless" in i):
+            ifname = i
+            ifname = ifname.split(':')[-1]
+            break
+    else:
+        ifname = "wlan0"
+        
+    com = Popen(["nmcli","--version"],stdout=PIPE)
+    com = com.communicate()[0].strip()
```
KeyWeeUsr: This will be a problem if you don't decode.

plyer/platforms/linux/wifi.py
```diff
@@ -80,15 +94,18 @@ def _connect(self, network, parameters):
...
+                self.ifname, network, cell, password
             )
         return result
 
     def _disconnect(self):
         '''
         Disconnect all the networks managed by Network manager.
         '''
-        return call(['nmcli', 'nm', 'enable', 'false'])
+        if(self.nmcliVersion > 98):
```
KeyWeeUsr: Also you are mixing #464 into this pull request...

PR #470, opened 2018-11-07
Wifi interface and nmcli's version fetching function added

xhimanshuz:
> Wifi interface name fetched if available else interface name will be 'wlan0'.
> Nmcli version checking function added to execute nmcli's command accordingly.
> Now disconnect all network with any nmcli's version.

### DIFF
--- a/plyer/platforms/linux/wifi.py
+++ b/plyer/platforms/linux/wifi.py
@@ -24,7 +24,7 @@ class LinuxWifi(Wifi):
 
     def getIfname(self):
         '''
-        Return wifi interface name if available else return "wlan0'
+        Return wifi interface name if available else return "wlan0'.
         '''
         com = Popen(["nmcli", "-t", "device"], stdout=PIPE)
         com = com.communicate()[0].decode().split("\n")
@@ -35,7 +35,7 @@ class LinuxWifi(Wifi):
 
     def getNmcliVersion(self):
         '''
-        Return nmcli version in Int to perform operation accordingly 
+        Return nmcli version in Int to perform operation accordingly.
         '''
         com = Popen(["nmcli", "--version"], stdout=PIPE)
         com = com.communicate()[0].decode("utf-8").strip()
--- a/plyer/platforms/linux/wifi.py
+++ b/plyer/platforms/linux/wifi.py
@@ -16,7 +16,30 @@ from subprocess import Popen, PIPE, call
 
 
 class LinuxWifi(Wifi):
+    def __init__(self):
+        self.ifname = self.getIfname()
+        self.nmcliVersion = self.getNmcliVersion()
+
     names = {}
+
+    def getIfname(self):
+        '''
+        Return wifi interface name if available else return "wlan0'
+        '''
+        com = Popen(["nmcli", "-t", "device"], stdout=PIPE)
+        com = com.communicate()[0].decode().split("\n")
+        for i in com:
+            if("wifi" in i):
+                return i.split(":")[0]
+        return 'wlan0'
+
+    def getNmcliVersion(self):
+        '''
+        Return nmcli version in Int to perform operation accordingly 
+        '''
+        com = Popen(["nmcli", "--version"], stdout=PIPE)
+        com = com.communicate()[0].decode("utf-8").strip()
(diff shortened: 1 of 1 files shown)


plyer/platforms/linux/wifi.py
```diff
@@ -16,8 +16,31 @@
...
     names = {}
 
+    def getIfname(self):
+        '''
+        Return wifi interface name if available else return "wlan0'.
+        '''
+        com = Popen(["nmcli", "-t", "device"], stdout=PIPE)
+        com = com.communicate()[0].decode().split("\n")
+        for i in com:
+            if("wifi" in i):
```
KeyWeeUsr: Python does not require brackets for conditions and it decreases the readability. Brackets in conditions are mostly used to explicitly mark an order of some *logical* operations for example:

```python
>>> 1 and (1 or 1) and 0
0
>>> 1 and 1 or 1 and 0
1
>>>
```

plyer/platforms/linux/wifi.py
```diff
@@ -80,15 +103,18 @@ def _connect(self, network, parameters):
...
+                self.ifname, network, cell, password
             )
         return result
 
     def _disconnect(self):
         '''
         Disconnect all the networks managed by Network manager.
         '''
-        return call(['nmcli', 'nm', 'enable', 'false'])
+        if(self.nmcliVersion > 98):
```
KeyWeeUsr: The same thing with the brackets as above.

plyer/platforms/linux/wifi.py
```diff
@@ -16,8 +16,31 @@
...
+                return i.split(":")[0]
+        return 'wlan0'
+
+    def getNmcliVersion(self):
+        '''
+        Return nmcli version in Int to perform operation accordingly.
+        '''
+        com = Popen(["nmcli", "--version"], stdout=PIPE)
+        com = com.communicate()[0].decode("utf-8").strip()
+        return int(com.split(" ")[-1].replace(".", ""))
```
KeyWeeUsr: In case you get a stupid value back from the process or you don't get *any* value back for whatever reason, you'll end up with `int('')` or `int('bad')` which will raise an error instead of continuing in the workflow normally where the commands for disconnect might actually work nevertheless. This must not fail. :)

plyer/platforms/linux/wifi.py
```diff
@@ -80,15 +108,18 @@ def _connect(self, network, parameters):
...
             )
         return result
 
     def _disconnect(self):
         '''
         Disconnect all the networks managed by Network manager.
         '''
-        return call(['nmcli', 'nm', 'enable', 'false'])
+        if self.nmcliVersion > 98:
+            return call(['nmcli', 'networking', 'off'])
```
KeyWeeUsr: I noticed the command is wrong. You are not ***disconnecting*** from the current network, you are disabling the whole networking interface.

plyer/platforms/linux/wifi.py
```diff
@@ -16,7 +16,35 @@
...
+        Return wifi interface name if available else return "wlan0'.
+        '''
+        try:
+            com = Popen(["nmcli", "-t", "device"], stdout=PIPE)
+            com = com.communicate()[0].decode().split("\n")
+            for i in com:
+                if "wifi" in i:
+                    return i.split(":")[0]
+            return 'wlan0'
+        except Exception:
```
KeyWeeUsr: Unless you are actually doing something with the exception, such as letting the user know what and why failed, the effect it the same as before.

plyer/platforms/linux/wifi.py
```diff
@@ -16,7 +16,35 @@
...
+
+    def _getNmcliVersion(self):
+        '''
+        Return nmcli version in Int to perform operation accordingly.
+        '''
+        try:
+            com = Popen(["nmcli", "--version"], stdout=PIPE)
+            com = com.communicate()[0].decode("utf-8").strip()
+            return int(com.split(" ")[-1].replace(".", ""))
+        except Exception:
```
KeyWeeUsr: Same here.

KeyWeeUsr:
Please open a new pull request with the reflected changes and conflicts resolved.

PR #471, opened 2018-11-10
thegrymek: Android speech recognition

KeyWeeUsr:
Recovering an open PR by @thegrymek with an `unknown repository`(deleted) patches.

> Here's android implementation for speech recognition.
>
> It still have issues with stopping because of error 'you must run it in main thread'. Rest work very well.
On sdk=9 I have error:client but documentation supports it. On sdk=21 it runs very well.
>
> best regards,
> Andrzej Grymkowski


KeyWeeUsr:
Requires any application that provides an [`android.speech.RecognitionService`](https://developer.android.com/reference/android/speech/RecognitionService) (no clue how to check for that). One of such applications is on a lot of devices preinstalled [Google (quick search box)](https://play.google.com/store/apps/details?id=com.google.android.googlequicksearchbox).

The API *prefers* offline recognition, but should be able to switch to online alternative in case you don't have a language package installed (internet permission necessary).

You can enable ***offline*** speech recognition this way (Android 8.1):
* open the `Settings` app
* choose `Language & Input`/`Language & Keyboard`/... (Samsung might include it in the `General` category)
* choose On-Screen keyboard or Voice search
* choose Google Keyboard
* Offline Speech recognition
* Download language package if you don't have one already

From my research, for some people the partial results are always empty. I've had no problems receiving them (start listening, say something, manual stop). The example app is written in a way that the standard listening (the service will stop your listener automatically) returns the result to the `results` field and if you start & stop all by yourself, then the field `partials` is filled. The rest depends on your implementation (e.g. if you want to display both partial + normal or show partial results while listening, etc etc).

PR #236, opened 2016-10-16
Update Wifi Module

manthansharma:
Add more functionality for wifi module and remove old bugs

PR #473, opened 2018-11-10
Add enable & disable WiFi for Linux and MacOS

KeyWeeUsr:
Rebased #236.

Enable and disable methods work fine on Linux Mint.


KeyWeeUsr:
Works fine on MacOS 10.13.16 too.

PR #474, opened 2018-11-11
Removed unnecessary grep dependency

Nephyx:
Removed unnecessary grep dependency from macOS bluetooth implementation, using python instead.


KeyWeeUsr:
Thanks! :dango:

Issue #451, opened 2018-10-14
Windows storagepath is missing application folder


KeyWeeUsr:
If we compare the name to let's say macOS alternative ([`NSApplicationDirectory`](https://developer.apple.com/documentation/foundation/nssearchpathdirectory/nsapplicationdirectory?language=objc)), we should most likely use `FOLDERID_Programs` from [`KNOWNFOLDERID`](https://docs.microsoft.com/en-us/windows/desktop/shell/knownfolderid).

ref: https://docs.microsoft.com/en-us/windows/desktop/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath

PR #517, opened 2019-10-03
Add Windows applications storage path

magnusvmt:
Added Windows applications storage path,
this will most likely point to C:\Program Files

Added Windows storage path tests.

Fixes issue #451


matham:
This returns only e.g. `C:\Program Files` and not both the x86 and x64 versions - even on 64 bit programs are stored in both, but still this is better than none.

PR #529, opened 2019-12-28
Add permission request to plyer GPS example

Zen-CODE:
Not certain here if including the android dependency in the *requirements* line is correct? How does one include OS specific dependencies in a cross-platform library? Sho


tshirtman:
thanks!

PR #537, opened 2020-01-31
fix-gps

chris-hamberg:
Fix for programmatically requesting permissions (required for Android API 23 and greater,) particularly for GPS on Android. TL;DR: this makes GPS work again, on Android!

In order for this fix to actually work some other parts of the Kivy framework also require minor patches. As they are specified in buildozer.spec the following values must be set:

android.api = 29
android.gradle_dependencies = 'com.google.android.gms:play-services-location:17.0.0'

It would be nice it could be possible to parse the permissions from buildozer.spec during build, and package it with a text file in the APK, or a simple AndroidManifest.xml parser that plyer could read on initialization. That way, the Manifest list in __init__ could properly be populated with all of the necessary permissions in one place, in case additional endpoints require this patch.

### DIFF
--- a/plyer/platforms/android/__init__.py
+++ b/plyer/platforms/android/__init__.py
@@ -3,6 +3,7 @@ from jnius import autoclass
 
 ANDROID_VERSION = autoclass('android.os.Build$VERSION')
 SDK_INT = ANDROID_VERSION.SDK_INT
+ActivityCompat = autoclass('androidx.core.app.ActivityCompat')
 
 try:
     from android import config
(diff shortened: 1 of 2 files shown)


Franccisco:
Hello! Thanks for the pull request.

I see you're trying to do something similar to #529, but in the constructor. To request permissions at runtime, you can use `android.permissions.request_permissions` in the `build` method, just like the pull request I mentioned

> It would be nice it could be possible to parse the permissions from buildozer.spec during build, and package it with a text file in the APK, or a simple AndroidManifest.xml parser that plyer could read on initialization. That way, the Manifest list in init could properly be populated with all of the necessary permissions in one place, in case additional endpoints require this patch.

`request_permissions` accepts a callback that is called when the permissions are accepted or denied.

In summary, it seems like a duplicate of #529, unless the example showed in there is not working.

PR #542, opened 2020-02-28
Add native iOS FileBrowser

Zen-CODE:
This PR needs the accompanying addition to kivy-ios. We cannot to all the work here from pjobjus as we need to call a non-class method `UIImagePNGRepresentation`. Otherwise, this could have been a stand-alone PR. 

https://github.com/kivy/kivy-ios/pull/428

Either way, these 2 PRs work together and allow plyer in invoke the native filechooser via the `NativeImagePicker` in the kivy-ios PR.


misl6:
Tested LGT.

PR #544, opened 2020-03-18
Switch to GitHub actions

Franccisco:
With this Pull Request, Travis CI and AppVeyor are both being replaced with GitHub Actions. The main reason is because with GitHub Actions you can run workflows in forks, without creating a pull request before.

Considerations:

* Python 2 is not supported.
* Secrets `PYPI.USERNAME`, `PYPI.PASSWORD`,  `COVERALLS_REPO_TOKEN`, and `GITHUB_RELEASE` must be set in the repository configuration.
* In order to keep consistency with the main kivy project, I created `sh` and `ps1` files with functions inside to run all the tasks.


.github/workflows/ci_osx.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - name: Install dependencies
+      run: |
+        source ci/ci_osx.sh
+        dependencies
+
+    - name: Style
+      run: |
+        source ci/ci_ubuntu.sh
+        style
```
matham: Style test should only be run on one os. Preferably Ubuntu. No need to run it on osx and windows as well.

.github/workflows/ci_ubuntu.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - name: Install dependencies
+      run: |
+        source ci/ci_ubuntu.sh
+        dependencies
+
+    - name: Style
+      run: |
+        source ci/ci_ubuntu.sh
+        style
```
matham: It would be good if the style test can be run as a separate job. Because it only needs to run on one python version. Also, style requires much less dependencies, so it should be much faster for that job to finish without people having to wait to see if there are PEP8 issues.

Finally, looking at the failures in the CI, it all has to do with imports. If this is a thing in plyer, maybe we can disable that check?

.github/workflows/plyer_deploy.yml
```diff
@@ -0,0 +1,30 @@
...
+        deployment_dependencies
+
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+
+      run: |
+        source ci/ci_ubuntu.sh
+        deploy
```
matham: Can we also add a upload to the github release page? Like in kivy.

.github/workflows/ci_ubuntu.yml
Franccisco: Respecting running the style test as a separate job -> it's done.

If I recall correctly, most of the import warning are because of the PY2 support, that will be removed soon. I don't know if the other imports are explicitly needed in the place they are.

By the way, considering that Kivy checks the style with flake8, should we do the same in plyer? Currently the checks are done with pylint and pycodestyle.

matham: >If I recall correctly, most of the import warnings are because of the PY2 support, that will be removed soon.

Is there a issue about dropping py2 support? If so we should make it clear that py2 support will be dropped after this upcoming release. And we can silence those py2 warning for now until it can be removed.

>By the way, considering that Kivy checks the style with flake8, should we do the same in plyer?

That sounds like a good idea.

.github/workflows/plyer_deploy.yml
```diff
@@ -0,0 +1,30 @@
+name: Deploy to PyPI
+
+on:
+  release:
+    types: [created]
```
matham: Are you sure that only tags are created and not e.g. branches? It may be better to run this simply `on: create`, and then have a check that it is a tag explicitly: `if: startsWith(github.ref, 'refs/tags/')` in the job root.

Kinda like at https://github.com/kivy-garden/graph/blob/master/.github/workflows/pythonapp.yml#L51. The workflow then is someone creates a tag locally or on gh. The CI runs and generates the sdist, the wheel and uploads to pypi and creates a draft release on github's release page and adds the files. Then we just need to publish the release draft and everything is set. This seems to work quite well.

.github/workflows/plyer_deploy.yml
Franccisco: Sure, I'll let you know when it's ready to review

.github/workflows/plyer_deploy.yml
Franccisco: As far as I know, the `on: release types: [created]` will run the job when the release is created in GitHub ([excluding the draft releases](https://help.github.com/en/actions/reference/events-that-trigger-workflows#release-event-release)). Using [this method](https://help.github.com/en/github/administering-a-repository/managing-releases-in-a-repository), a tag is required. I don't see in documentation a place where it says it creates a branch or something that is not a tag.

Maybe you do not prefer this method. If I understood, you want plyer_deploy to run just with the creation of a tag, creating an artifact and a draft release, don't you? If so, I can work on it.

matham: >Maybe you do not prefer this method. If I understood, you want plyer_deploy to run just with the creation of a tag, creating an artifact and a draft release, don't you? If so, I can work on it.

Yes, I'd prefer running it with a tag as that is in line with our overall workflow. Thanks

Franccisco: I just added those. Can you review the changes please?

.github/workflows/plyer_deploy.yml
```diff
@@ -0,0 +1,46 @@
...
+  deploy:
+    if: startsWith(github.ref, 'refs/tags/')
+
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.7.x'
```
matham: Why `3.7.x`? Isn't `3.x` better?

.github/workflows/plyer_deploy.yml
```diff
@@ -0,0 +1,46 @@
...
+        draft: true
+
+    - name: Build and publish to PyPI
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+
+      run: |
+        source ci/ci_ubuntu.sh
+        deploy
```
matham: You are running the deploy step after the artifacts and release upload. I think this needs to happen first.

So the order should be: generate deploy artifacts, then upload artifact to job, then release upload, and finally pypi upload. Pypi upload always needs to be last because that is the most likely to fail, so that's why the order is artifact (least likely to fail) and release page (slightly more likely to fail compared to artifact, but less so than pypi) and then pypi.

ci/ci_ubuntu.sh
```diff
@@ -0,0 +1,74 @@
...
+
+  # submit coverage report from tests to coveralls.io
+  # requires: REPO_TOKEN, SERVICE_NAME, JOB_ID, PULL_REQUEST
+  coveralls || true
+}
+
+deploy()
+{
+  # deploy to PyPI
+  python setup.py sdist bdist_wheel
```
matham: For `bdist_wheel` we need to specify the `--universal` flag. So the sdist and bdist would need to be on separate lines.

matham: Also, you should separate generating artifact from pypi upload as mentioned earlier.

.github/workflows/plyer_deploy.yml
Franccisco: I agree.

.github/workflows/plyer_deploy.yml
Franccisco: I did it here https://github.com/kivy/plyer/pull/544/commits/1159d9270f668d3b1084e327902e8556be959f9e

Franccisco:
@matham One more thing: I included the coverage upload to coveralls.io in ci_ubuntu, but as I configured the tests job to run with py 3.6, 3.7 and 3.8, then the upload is triggered three times. Maybe I can fix it adding an if statement, but in Kivy the tests are only checked using Python 3.x, so i did the same. If you have any suggestion about that, feel free to share

matham:
>One more thing: I included the coverage upload to coveralls.io in ci_ubuntu, but as I configured the tests job to run with py 3.6, 3.7 and 3.8, then the upload is triggered three times. Maybe I can fix it adding an if statement, but in Kivy the tests are only checked using Python 3.x, so i did the same.

I'm ok with either way.

matham:
Thanks for all your work!! I added all the secret variables to the settings.

I guess we'll find out at the next release if it worked.

Issue #541, opened 2020-02-24
battery.status isCharging always shows false in WINDOWS


irm19:
Hello, 

I have noticed that battery.status['isCharging'] was false in my computer while it was connected. 

The problem is related to line 22 in file https://github.com/kivy/plyer/blob/master/plyer/platforms/win/battery.py

        status["isCharging"] = query["BatteryFlag"] == 8

The problem is that "BatteryFlag" is actually a byte composed of multiple bits, see.. 
https://docs.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-system_power_status
so if multiple conditions apply, previous value would tend to be false... (in my case it was 'low battery' and 'charging') :-)

So my proposal is to consider somethin like 

         status["isCharging"] = query["BatteryFlag"] & 8 > 0

However, that would still generate issues when BatteryFlag is 255 (Unknown)... so maybe...

         status["isCharging"] = query["BatteryFlag"] !=255 and query["BatteryFlag"] & 8 > 0

There are multiple options to handle this case... :-)

Keep up!!

Franccisco:
Hello! I suggest you to open a pull request with these changes, so they can be tested and eventually merged.

irm19:
Associated pull request #546.
For some unknown reason, it is not passing the "Ubuntu" styles. I assume something else must be wrong in the ubuntu construction, as the file modified is in the Windows side.

Zen-CODE:
Closed by the above merge. Thanks

PR #546, opened 2020-03-23
Modification of Status isCharging in windows

irm19:
Modification to adapt that the source of information is a flag that includes multiple conditions. See.
https://docs.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-system_power_status

Associated to issue #541

### DIFF
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -23,7 +23,7 @@ class WinBattery(Battery):
             return status
 
         status["isCharging"] = (query["BatteryFlag"] != UNKNOWN_STATUS) and \
-                               (query["BatteryFlag"] & NO_BATTERY > 0) and \
+                               (query["BatteryFlag"] & NO_BATTERY == 0) and \
                                (query["BatteryFlag"] & CHARGING > 0)
         status["percentage"] = query["BatteryLifePercent"]
 
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -13,7 +13,7 @@ class WinBattery(Battery):
 
     def _get_state(self):
         CHARGING = 8
-        NO_SYS_BATTERY = 128
+        NO_BATTERY = 128
         UNKNOWN_STATUS = 255
         status = {"isCharging": None, "percentage": None}
 
@@ -23,7 +23,7 @@ class WinBattery(Battery):
             return status
 
         status["isCharging"] = (query["BatteryFlag"] != UNKNOWN_STATUS) and \
-                               (query["BatteryFlag"] & NO_SYS_BATTERY > 0) and \
+                               (query["BatteryFlag"] & NO_BATTERY > 0) and \
                                (query["BatteryFlag"] & CHARGING > 0)
         status["percentage"] = query["BatteryLifePercent"]
 
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -4,6 +4,7 @@ Module of Windows API for plyer.battery.
 
 from plyer.platforms.win.libs.batterystatus import battery_status
 from plyer.facades import Battery
+from ctypes.wintypes import BYTE
 
 
 class WinBattery(Battery):
@@ -12,8 +13,8 @@ class WinBattery(Battery):
     '''
 
     def _get_state(self):
-        CHARGING = 8
-        UNKNOWN_STATUS = -1
+        CHARGING = BYTE(8).value
+        UNKNOWN_STATUS = BYTE(255).value
         status = {"isCharging": None, "percentage": None}
 
         query = battery_status()
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -19,7 +19,8 @@ class WinBattery(Battery):
         if not query:
             return status
 
-        status["isCharging"] = query["BatteryFlag"] !=255 and query["BatteryFlag"] & 8 > 0
+        status["isCharging"] = query["BatteryFlag"] !=255 and 
+                               query["BatteryFlag"] & 8 > 0
         status["percentage"] = query["BatteryLifePercent"]
 
         return status
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -12,11 +12,9 @@ class WinBattery(Battery):
     '''
 
     def _get_state(self):
-        # Flags values
         CHARGING = 8
-        NO_SYSTEM_BATTERY = 128
+        NO_SYS_BATTERY = 128
         UNKNOWN_STATUS = 255
-        
         status = {"isCharging": None, "percentage": None}
 
         query = battery_status()
@@ -25,7 +23,7 @@ class WinBattery(Battery):
             return status
 
         status["isCharging"] = (query["BatteryFlag"] != UNKNOWN_STATUS) and \
-                               (query["BatteryFlag"] & NO_SYSTEM_BATTERY > 0) and \
+                               (query["BatteryFlag"] & NO_SYS_BATTERY > 0) and \
                                (query["BatteryFlag"] & CHARGING > 0)
         status["percentage"] = query["BatteryLifePercent"]
 
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -12,6 +12,12 @@ class WinBattery(Battery):
     '''
 
     def _get_state(self):
+        
+        # Flags values 
+        CHARGING = 8
+        NO_SYSTEM_BATTERY = 128
+        UNKNOWN_STATUS = 255
+        
         status = {"isCharging": None, "percentage": None}
 
         query = battery_status()
@@ -19,8 +25,9 @@ class WinBattery(Battery):
         if not query:
             return status
 
-        status["isCharging"] = (query["BatteryFlag"] != 255) and \
-                               (query["BatteryFlag"] & 8 > 0)
+        status["isCharging"] = (query["BatteryFlag"] != UNKNOWN_STATUS) and \
+                               (query["BatteryFlag"] & NO_SYSTEM_BATTERY > 0) and \
+                               (query["BatteryFlag"] & CHARGING > 0)
         status["percentage"] = query["BatteryLifePercent"]
 
         return status
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -12,8 +12,7 @@ class WinBattery(Battery):
     '''
 
     def _get_state(self):
-        
-        # Flags values 
+        # Flags values
         CHARGING = 8
         NO_SYSTEM_BATTERY = 128
         UNKNOWN_STATUS = 255
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -19,8 +19,8 @@ class WinBattery(Battery):
         if not query:
             return status
 
-        status["isCharging"] = query["BatteryFlag"] !=255 and \
-                               query["BatteryFlag"] & 8 > 0
+        status["isCharging"] = (query["BatteryFlag"] != 255)  and  \
+                               (query["BatteryFlag"] & 8 > 0)
         status["percentage"] = query["BatteryLifePercent"]
 
         return status
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -19,7 +19,7 @@ class WinBattery(Battery):
         if not query:
             return status
 
-        status["isCharging"] = query["BatteryFlag"] !=255 and 
+        status["isCharging"] = query["BatteryFlag"] !=255 and \
                                query["BatteryFlag"] & 8 > 0
         status["percentage"] = query["BatteryLifePercent"]
 
--- a/plyer/platforms/win/battery.py
+++ b/plyer/platforms/win/battery.py
@@ -19,7 +19,7 @@ class WinBattery(Battery):
         if not query:
             return status
 
-        status["isCharging"] = (query["BatteryFlag"] != 255)  and  \
+        status["isCharging"] = (query["BatteryFlag"] != 255) and \
                                (query["BatteryFlag"] & 8 > 0)
         status["percentage"] = query["BatteryLifePercent"]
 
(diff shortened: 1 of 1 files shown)


Franccisco:
> For some unknown reason, it is not passing the "Ubuntu" styles. I assume something else must be wrong in the ubuntu construction, as the file modified is in the Windows side.

The style job in Ubuntu performs checks with flake8. Currently there are many warnings, and many of them are not related with `battery.py`, so that build will fail. You can check the logs of the `Style` step if the file you are modifying has warnings:

```
./plyer/platforms/win/battery.py:22:55: E225 missing whitespace around operator
        status["isCharging"] = query["BatteryFlag"] !=255 and \
                                                      ^
./plyer/platforms/win/battery.py:23:32: E127 continuation line over-indented for visual indent
                               query["BatteryFlag"] & 8 > 0
                               ^
```

irm19:
Code modified to avoid the previous issue. The only thing missing regarding battery is related to the examples files, and I guess these are generic issues associated to all the existing data. Thanks Franccisco 

./examples/battery/main.py:3:1: F401 'kivy.uix.button.Button' imported but unused
^
./examples/battery/main.py:5:1: F401 'kivy.properties.StringProperty' imported but unused
from kivy.properties import StringProperty, ObjectProperty
^

Franccisco:
You're welcome!

I confirm this pull request is working as expected. Just one more thing:

Instead of having `255` and `8` as magic numbers, can you define constants for them? For example:

```py
CHARGING = 8
UNKNOWN_STATUS = 255
```

Franccisco:
> The only thing missing regarding battery is related to the examples files, and I guess these are generic issues associated to all the existing data.

You're right. Don't worry about those warnings, they are going to be fixed in another pull request.

irm19:
Yes, it is possible, but I would limit the scope where to put them. Actually reviewing the flags in windows, there is also another value that may affect the overall behaviour... `128 | No system battery`. I'm not sure, but if there is no battery it should never be charging... :-)

plyer/platforms/win/battery.py
```diff
@@ -12,14 +12,19 @@ class WinBattery(Battery):
...
         status = {"isCharging": None, "percentage": None}
 
         query = battery_status()
 
         if not query:
             return status
 
-        status["isCharging"] = query["BatteryFlag"] == 8
+        status["isCharging"] = (query["BatteryFlag"] != UNKNOWN_STATUS) and \
+                               (query["BatteryFlag"] & NO_BATTERY > 0) and \
```
Franccisco: I would not add this check. Consider the case where `query["BatteryFlag"]` is in CHARGING and High battery level, then it takes the value of 9 and `query["BatteryFlag"] & NO_BATTERY > 0` is `False`, then `status["isCharging"]` will be `False` too.

If I'm not wrong, `NO_BATTERY` can't be in conjunction with another flag, so in the case `query["BatteryFlag"]` is 128, `(query["BatteryFlag"] != UNKNOWN_STATUS) and (query["BatteryFlag"] & CHARGING > 0)` will be `False`.

So I suggest you to remove this and the `NO_BATTERY` constant.
```suggestion
```

plyer/platforms/win/battery.py
```diff
@@ -12,14 +12,19 @@ class WinBattery(Battery):
     '''
 
     def _get_state(self):
+        CHARGING = 8
+        NO_BATTERY = 128
```
Franccisco: Read what I put below.
```suggestion
```

irm19:
Well, I have updated the code with the additional safeguard. The flags have been added as constants at the beggining of the function, and being honest, I'm starting to hate the warnings of the limitation of 79 characters per line. Not sure how many people still have that limitation in the screen nowadays, but makes the code less readable for sure...

Franccisco:
> Well, I have updated the code with the additional safeguard. The flags have been added as constants at the beggining of the function, and being honest, I'm starting to hate the warnings of the limitation of 79 characters per line. Not sure how many people still have that limitation in the screen nowadays, but makes the code less readable for sure...

I just left you a review. About the characters limitation... I know it might be annoying sometimes, we just moved to flake8 and for now it is checking everything with the default settings. Despite that, you're doing a great job in this pull request and I really appreciate it!

plyer/platforms/win/battery.py
irm19: We crossed.... :-) 
Let's see... in the documentation this is what is mentioned for `BatteryFlag`

Value | Meaning
1 | High—the battery capacity is at more than 66 percent
2 | Low—the battery capacity is at less than 33 percent
4 | Critical—the battery capacity is at less than five percent
8 | Charging
128 | No system battery
255 | Unknown status—unable to read the battery flag information

Per my understanding:
- UNKNOWN STATUS = 1111 1111
- NO SYSTEM BATTERY = 0100 0000
- CHARGING = 0000 1000
- CRITICAL = 0000 0100
- LOW = 0000 0010
- HIGH = 0000 0001

So...
- UNKNOWN STATUS requires all bits to be 1, so it must be treated combined as an speciall case 
- NO SYSTEM BATTERY is actually a bit flag, and if things are reasonable, it should never be used when CHARGING, CRITICAL, LOW or HIGH. I mean, if you rely on windows working correctly, because if there is no battery... it cannot be on any of those states,... right?
- CHARGING is a flag, and can be simultaneously with CRITICAL, LOW or HIGH, 
- CRITICAL, LOW or HIGH should be mutually exclusive... (except if it is just changing?)

At least that is what I understand should happen if things should work as expected isn't it?  

But this is about if not having a battery can allow to charge or have other level states. I'm trying to think about strange cases, and I can only think about the possibility of having multiple batteries, and one of them missing... :-). If so, those flags could be independent except if in windows the structure of SYSTEM_POWER_STATUS is used per battery. Honestly,  I don't know, and it is going to be slightly difficult to get a 2 battery laptop... :-)

Still, the real point is to know  what happens with the flags when we remove the battery... so... Practical approach... Test and decide... I'll check it in my laptop and let you know...

Franccisco: > CRITICAL, LOW or HIGH should be mutually exclusive... (except if it is just changing?)
What if the battery percentage is at 4%? Both CRITICAL and LOW are matching the condition.

> Test and decide... I'll check it in my laptop and let you know...
I just tested your last modifications and I always get `False`. I'm getting the expected result with what I suggested.

irm19: > What if the battery percentage is at 4%? Both CRITICAL and LOW are matching the condition.

On this I agree, I wrote it too fast. Low condition would still be applicable if Critical condition is happening... actually, I just checked it (see nearly at the end).

> I just tested your last modifications and status["isCharging"] is False. I'm getting the expected result with what I suggested.

Ummm, I have just noticed that my code was wrong... as the condition... 
             `(query["BatteryFlag"] & NO_BATTERY > 0)`
would be checking that there is no battery... it should have really been that condition not being fulfilled... so either
              `(query["BatteryFlag"] & NO_BATTERY == 0)`
or  
              ` not (query["BatteryFlag"] & NO_BATTERY > 0)`

**But, being honest, I think even with that there is something else going on behind scenes... at least in my Win10 Home 64bits.**

1. With the battery full, as expected BatteryFlag is equal to 1 (High). It was not loading, so it seemed fine. 

```
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 0,
 'BatteryFlag': 1,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': 100,
 'BatteryLifeTime': 14020,
 'Reserved1': 0}
```

2. I removed the battery keeping the power... and... Here it comes a surprise!!! 
### **-128** 
Definetly I was not expecting this negative value... 

```
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 1,
 'BatteryFlag': -128,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': -1,
 'BatteryLifeTime': 4294967295,
 'Reserved1': 0}
```

3. I did put the battery again... and even if the icon in windows looked like charging...  it was reporting as it wasn't... (it could be though)
```
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 1,
 'BatteryFlag': 1,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': 99,
 'BatteryLifeTime': 4294967295,
 'Reserved1': 0}
```

4. So I have let discharge a bit more the battery...  and now it seems reasonable... charging and high
```
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 1,
 'BatteryFlag': 9,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': 96,
 'BatteryLifeTime': 4294967295,
 'Reserved1': 0}
```

5. Let's see what happens when we reduce a lot the battery.... at 66%
```
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 0,
 'BatteryFlag': 0,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': 66,
 'BatteryLifeTime': 6699,
 'Reserved1': 0}
```
6. At 30%
```
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 0,
 'BatteryFlag': 2,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': 30,
 'BatteryLifeTime': 1997,
 'Reserved1': 0}
```


7. At 5%, 4%, and charging...  (This demonstrate you were right about LOW and CRITICAL)
```
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 0,
 'BatteryFlag': 2,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': 5,
 'BatteryLifeTime': 418,
 'Reserved1': 0}
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 0,
 'BatteryFlag': 6,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': 4,
 'BatteryLifeTime': 462,
 'Reserved1': 0}
[Dbg]>>> query = battery_status()
[Dbg]>>> query
{'ACLineStatus': 1,
 'BatteryFlag': 14,
 'BatteryFullLifeTime': 4294967295,
 'BatteryLifePercent': 4,
 'BatteryLifeTime': 4294967295,
 'Reserved1': 0}

```

So now my question would be... What value would actually have UNKNOWN STATUS? If it is -255 after the surprise I got with the NO SYSTEM BATTERY, that code may also need additional investigation... and verification... (and we can start even guessing how to trigger it).  For the NO SYSTEM BATTERY case, we can make a quick adjustment because   in python `-128 & 128` is equal to 128... so `(query["BatteryFlag"] & NO_BATTERY == 0)` would still be applicable... So for now I would go for that solution, till we actually are able to know te other case "UNKNOWN STATUS"...

I know all this is a mess... :-)

irm19: Just another comment, I think all this difference in the value is related on how it is been read a BYTE from this structure 

typedef struct _SYSTEM_POWER_STATUS {
  BYTE  ACLineStatus;
  BYTE  BatteryFlag;
  BYTE  BatteryLifePercent;
  BYTE  SystemStatusFlag;
  DWORD BatteryLifeTime;
  DWORD BatteryFullLifeTime;
} SYSTEM_POWER_STATUS, *LPSYSTEM_POWER_STATUS;

My guess is that this BYTE that is suposed to be unsigned is being read as a complementary 2 value. in that case 128 would be read as -128, and 255 would be read as -1, but not sure at all about the procedure used to read from the C structure... I'll try to take a look at it at some point.

Franccisco: Thanks for testing!

> Let's see what happens when we reduce a lot the battery.... at 66%

This is expected as it appears in the documentation.

> My guess is that this BYTE that is suposed to be unsigned is being read as a complementary 2 value. in that case 128 would be read as -128, and 255 would be read as -1, but not sure at all about the procedure used to read from the C structure... I'll try to take a look at it at some point.

The BYTE type is extracted from ctypes.wintype:
https://github.com/kivy/plyer/blob/6a3d81e5517a8aba7c450d421243c95846d4a2ca/plyer/platforms/win/libs/win_api_defs.py#L13-L14

https://github.com/kivy/plyer/blob/6a3d81e5517a8aba7c450d421243c95846d4a2ca/plyer/platforms/win/libs/win_api_defs.py#L192

It is actually an alias for ctypes.c_byte:

https://github.com/python/cpython/blob/34a49aa3e4d023b5f9e9029f4f1ec68f1a8a8120/Lib/ctypes/wintypes.py#L4

And ctypes.c_byte is a [C signed char datatype](https://docs.python.org/3/library/ctypes.html?highlight=c_int#ctypes.c_byte).

By the way, I just found an outdated definition in win_api_defs.py, but we can fix it in another pull request.

irm19: Ok, so it is a matter of of convention as I expected. 

| Value | Meaning |Binary| Read as...|
|---|---|---|---|
|1 | High—the battery capacity is at more than 66 percent |0000 0001| 1|
|2 | Low—the battery capacity is at less than 33 percent | 0000 0010| 2|
|4 | Critical—the battery capacity is at less than five percent |0000 0100| 4|
|8 | Charging |0000 1000| 8|
|128 | No system battery |1000 0000| -128|
|255 | Unknown status—unable to read the battery flag information |1111 1111 | -1|

And as we have seen that  "No system battery" is not actually used simultaneously with the other  High, Low, Critical and Charging (and Charging would read false), my suggestion would be to keep it simple and only check  for Unknown Status (with value -1).  That's what I'm going to leave. Just let me know if you prefer to change something else.

Franccisco: You're right. Actually, those are the returned values if you do:

```py
from ctypes.wintypes import BYTE

BYTE(128).value
```

it returns -128, and with 255 the returned value is -1.

So, there are two things you can do:

1. Add a comment explaining why you set UNKNOWN_STATUS to -1 instead of 255.
2. Use BYTE.value to define the CHARGING and UNKNOWN_STATUS, like:

```py
UNKNOWN_STATUS = BYTE(255).value
```

With that, it will be easier to understand and therefore fix any future issue.

irm19: Done with the aproach of BYTE(XXX).value in all cases.

PR #547, opened 2020-03-28
Fix linter warnings

Franccisco:
This does not include the py2 warnings as py2 support is being dropped in this plyer version.

### DIFF
--- a/plyer/platforms/ios/flash.py
+++ b/plyer/platforms/ios/flash.py
@@ -3,7 +3,7 @@
 Flash
 -----
 """
-
+from plyer.facades import Flash
 from pyobjus import autoclass
 
 NSString = autoclass("NSString")
--- a/plyer/platforms/win/libs/wifi_defs.py
+++ b/plyer/platforms/win/libs/wifi_defs.py
@@ -5,8 +5,24 @@ https://msdn.microsoft.com/en-us/library/windows/desktop \
 /ms705945%28v=vs.85%29.aspx
 '''
 
-from ctypes import *
-from ctypes.wintypes import *
+from ctypes import (
+    POINTER,
+    FormatError,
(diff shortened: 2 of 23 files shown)


plyer/platforms/ios/flash.py
```diff
@@ -3,7 +3,7 @@
 Flash
 -----
 """
-
+from plyer.facades import Flash
```
Zen-CODE: This should probably be:

    from plyer.facades.flash import Flash

plyer/platforms/win/libs/wifi_defs.py
```diff
@@ -5,8 +5,24 @@
 /ms705945%28v=vs.85%29.aspx
 '''
 
-from ctypes import *
-from ctypes.wintypes import *
+from ctypes import (
```
Zen-CODE: Are we sure these are all the used imports?

plyer/platforms/ios/flash.py
Franccisco: I think it's right to do `form plyer.facades import Flash`, because your suggestion is already defined here:

https://github.com/kivy/plyer/blob/6a3d81e5517a8aba7c450d421243c95846d4a2ca/plyer/facades/__init__.py#L26

plyer/platforms/win/libs/wifi_defs.py
Franccisco: Sure. In addition, linter is not complaining about undefined names.

plyer/platforms/ios/flash.py
Zen-CODE: Right you are! :-)

Zen-CODE:
Thanks +1

PR #548, opened 2020-03-29
Remove unused linters

Franccisco:
As pylint and pycodestyle were replaced for flake8 in #544, this pull request removes them from the .py files.

### DIFF
--- a/plyer/__init__.py
+++ b/plyer/__init__.py
@@ -19,7 +19,6 @@ __version__ = '2.0.0.dev0'
 from plyer import facades
 from plyer.utils import Proxy
 
-# pylint: disable=invalid-name
 #: Accelerometer proxy to :class:`plyer.facades.Accelerometer`
 accelerometer = Proxy('accelerometer', facades.Accelerometer)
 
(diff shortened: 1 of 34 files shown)


Franccisco:
Hello @Zen-CODE, this is ready to review.

AndreMiras:
Oh and I didn't see, but actually the flake8 is also failing now, so what I would recommend is add global exception using the config file to keep the build green and then start addressing the ones we want to address. Problem with non-green build is then people take less care because "the build is red anyway" so things start to pile

Zen-CODE:
@AndreMiras Agreed. Squashing for the win!
On point 2, also agreed. But I was secretly hoping @Franccisco  would tackle those issues next. A lot are xrange python2 legacy warnings, and those should really be changed to range. They are valid warnings..

Zen-CODE:
@Franccisco Thank you. It is appreciated. +1

Franccisco:
I also agree with AndreMiras. Doing a squash is ideal.
You're welcome!

Issue #449, opened 2018-10-14
macOS notification NSUserNotificationCenter is deprecated + missing Info.plist


KeyWeeUsr:
https://github.com/kivy/plyer/blob/ee44e6e786f99ba5d620fb66ce558c1bc50ffc54/plyer/platforms/macosx/notification.py#L46

and also `defaultUserNotificationCenter()` returns `None` due to missing instructions for macOS
throws a `NoneType` not having a `setDelegate_` attribute.

For other than `None` result of `defaultUserNotificationCenter()` there has to be a `Info.plist` for the application.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>CFBundleIdentifier</key>
    <string>org.python.python</string>
</dict>
</plist>
```

ref: https://github.com/jaredks/rumps/issues/9
ref: https://github.com/jaredks/rumps/pull/36/files

akshayaurora:
This is no longer a issue with vents with python 3.9+ from python.org + venvs and Monterey. Re-open if you cab reproduce with new versions.

PR #551, opened 2020-04-06
Notification OS X :  use Apple Script Notification

wasimafser:
Depreceation of NSUserNotificationCenter causes problems with Mojave+ i guess

in reference with #449

### DIFF
--- a/plyer/platforms/macosx/notification.py
+++ b/plyer/platforms/macosx/notification.py
@@ -4,19 +4,7 @@ Module of MacOS API for plyer.notification.
 
 from plyer.facades import Notification
 
-from pyobjus import (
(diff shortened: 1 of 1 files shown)


beauwilliams (approved):
This works perfectly for myself. Reduces dependency. Would like to see this merged to main.

akshayaurora:
Current implementation has no issues on Monterey using venv. Using Python.org -> python 3.9+ .

Would be open for having this as a alternative fallback if original fails.

yeti9990:
Current implementation fails for me on:
```
Mac Studio M1
Mac Sonoma 14.2.1
plyer                2.1.0
pyobjus              1.2.3
```
This worked closed PR implementation worked for me.

masterfelu:
this works

PR #557, opened 2020-04-19
Getting Unique ID on Mac OS

vonkuptschino:


Franccisco:
Hello

After checking your changes, I don't think this pull request is related
to MacOS only. Could you elaborate?

vonkuptschino:
Hello,
actually your words make sense. I created PR before I tested it on any other platforms than Mac OS. Now I should say that it also works for Android and Windows. Unfortunately, I have no opportunity to elaborate on it for iOS.

akshayaurora:
There is no rational provided for this change,  please re-open with a proper explanation of what this pr does and why it's needed.

PR #563, opened 2020-05-17
add installation section in README.md

tshirtman:


README.md
```diff
@@ -59,6 +59,10 @@ implement the api in the easiest way, depending on the current platform.
 | Vibrator                       | ✔       | ✔   |         |      |       |
 | Wifi                           |         |     | ✔       | ✔    | ✔     |
 
+## Installation
+
+To use on desktop: `pip install plyer`
+To use in python-for-android/kivy-ios: add `plyer` to your requirements if needed (but it's probably already there).
```
Franccisco, line 65: > if needed (but it's probably already there).

Is this really necessary? As far as I know, `plyer` is not installed by
default in python-for-android/kivy-ios.

Franccisco, line 65: > if needed (but it's probably already there).

Is this really necessary? As far as I know, plyer is not installed by
default in either python-for-android or kivy-ios.

tshirtman: you are right, that's actually misleading, i'll remove that part.

tshirtman, line 65:
```suggestion
To use in python-for-android/kivy-ios: add `plyer` to your requirements if needed.
```

PR #565, opened 2020-06-30
Fixing crash on MacOSX filechooser

### LINKED ISSUE #524: Filechooser on mac: using path, crashes python
MacOS 10.14.6

using the path parameter causes python to crash (not just the app).
Using the use_extensions with path causes an uncaught exception.

Test case:
```
from pathlib import Path

from kivy.app import App
from kivy.lang.builder import Builder
from plyer import filechooser

kv = '''
BoxLayout:
    orientation: 'vertical'
    Button:
        text: 'Test: Open, filters=["*.pdf"]; this will crash python'
        on_release: app.ft.open(filters=['*.pdf'])
    Button:
        text: 'Test: Open, filters=["*.pdf"] use_extensions=True; throws exception '
        on_release: app.ft.open(filters=['*.pdf'], use_extensions=True)
    Button:
        text: 'Test: Open, filters=[] this will work as expected'
        on_release: app.ft.open()
    Label:
        text: 'plyer.filechooser.open_file() with a filter will crash on mac'
'''


class FileTest:
    def __init__(self):
        self.file_name = ''

    def open(self, **kwargs):
        filechooser.open_file(title='Open Patch File',
                              on_selection=self._open_selection, **kwargs)

    def _open_selection(self, selection):
        try:
            self.file_name = Path(selection[0]).name
        except (ValueError, IndexError):  # The user did not select a file
            print('open canceled')
        else:
            print(f'open file: {self.file_name}')


class PlyerFilechooserTestApp(App):
    ft = FileTest()

    def build(self):
        return Builder.load_string(kv)


if __name__ == '__main__':
    PlyerFilechooserTestApp().run()

Mulugruntz:
Closes #524.

### DIFF
--- a/plyer/platforms/macosx/filechooser.py
+++ b/plyer/platforms/macosx/filechooser.py
@@ -85,14 +85,14 @@ class MacFileChooser:
                         pystr = f.strip().split("*")[-1].split(".")[-1]
                     filthies.append(objc_str(pystr))
                 else:
-                    for _ in f[1:]:
+                    for s in f[1:]:
                         if not self.use_extensions:
-                            if f.strip().endswith("*"):
+                            if s.strip().endswith("*"):
                                 continue
-                            pystr = f.strip().split("*")[-1].split(".")[-1]
+                            pystr = s.strip().split("*")[-1].split(".")[-1]
                         filthies.append(objc_str(pystr))
 
-            ftypes_arr = objc_arr(filthies)
+            ftypes_arr = objc_arr(*filthies)
             panel.setAllowedFileTypes_(ftypes_arr)
             panel.setAllowsOtherFileTypes_(not self.use_extensions)
 


Mulugruntz:
@matham do you mean checking if it's a `str`-like before doing `.strip()`?

matham:
No, if `self.use_extensions` is True, all the code will break (including above your changes) because `pystr` would be undefined. Unless I'm overlooking something.

Mulugruntz:
You're right :-|

I'm not entirely sure where the block from `if self.use_extensions` should end but if we're assuming it's just after the `continue`, then yes there's an indent issue there :-/

matham:
I thought that the meaning `use_extensions` is that we don't need to do any conversions so basically it would just be `pystr = s` in the else block and the same (`pystr = f`) above that. But I could be wrong so we could just leave it. I suspect no one ever set `use_extensions` to True :D

PR #568, opened 2020-07-01
Fix uninitialized variable in MacOSX filechooser.

### LINKED ISSUE #566: Filter variable may be uninitialized in MacOSX filechooser
Variable `pystr` may not be initialized
https://github.com/kivy/plyer/blob/dec02d7/plyer/platforms/macosx/filechooser.py#L85

Moreover, there's code duplication so this bug is also duplicated.

Mulugruntz:
Closes #566.

### DIFF
--- a/plyer/platforms/macosx/filechooser.py
+++ b/plyer/platforms/macosx/filechooser.py
@@ -75,22 +75,17 @@ class MacFileChooser:
         # Mac OS X does not support wildcards unlike the other platforms.
         # This tries to convert wildcards to "extensions" when possible,
         # ans sets the panel to also allow other file types, just to be safe.
-        if len(self.filters) > 0:
+        if self.filters:
             filthies = []
             for f in self.filters:
                 if type(f) == str:
+                    f = (None, f)
+                for s in f[1:]:
                     if not self.use_extensions:
-                        if f.strip().endswith("*"):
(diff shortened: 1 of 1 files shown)


Mulugruntz:
Any update?

plyer/platforms/macosx/filechooser.py
```diff
@@ -75,22 +75,17 @@ def run(self):
...
             for f in self.filters:
                 if type(f) == str:
+                    f = (None, f)
+                for s in f[1:]:
                     if not self.use_extensions:
-                        if f.strip().endswith("*"):
+                        if s.strip().endswith("*"):
                             continue
-                        pystr = f.strip().split("*")[-1].split(".")[-1]
+                    pystr = s.strip().split("*")[-1].split(".")[-1]
```
tshirtman, line 87: this line was previously conditional to `not self.use_extensions` is the change voluntary?

Mulugruntz: Yes, that's actually the main fix to the bug (`pystr` may not be initialized). I believe it was not meant to be indented in the first place, as you could very well have `filters=["*"]`, which is valid... but will crash on MacOS.

The rest of the changes is (mostly) to remove code duplication and to fix it once instead of twice.

PR #594, opened 2020-12-12
Correctly remove missing Bluetooth from matrix

xloem:
The support for Bluetooth in the repository is a nonbehaving stub. It shouldn't be marked as supported, as only hardware enabled status is implemented and no communication with devices.
It would be good if a label were added asking for help implementing it.


akshayaurora:
So we should separate Bluetooth into Bluetooth->`FeatureX` specifying which use case is supported and which isn't

rshah713 (approved):

PR #612, opened 2021-03-30
DestroyWindow before UnregisterClass

Inso-m-niaC:
Fixes the issue where systray notification icons would linger until hovered over with mouse pointer.


Inso-m-niaC:
The problem lies elsewhere, not plyer. Closing

PR #616, opened 2021-04-27
Fix `UnicodeDecodeError` in `Win32FileChooser` class in `run` function

Neizvestnyj:
Previously, when using the Latin alphabet, an error occurred:
```
self.selection = [str(get_path(pidl).decode('utf-8'))]
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd0 in position 24: invalid continuation byte
```
So that this error no longer occurs, the encoding was changed to `Windows-1251`, and such errors now do not lead to the termination of the program with an error.

### DIFF
--- a/plyer/platforms/win/filechooser.py
+++ b/plyer/platforms/win/filechooser.py
@@ -136,7 +136,7 @@ class Win32FileChooser:
                 # pidl is None when nothing is selected
                 # and e.g. the dialog is closed afterwards with Cancel
                 if pidl:
-                    self.selection = [str(get_path(pidl).decode('utf-8'))]
+                    self.selection = [str(get_path(pidl).decode('Windows-1251', errors='ignore'))]
 
         except (RuntimeError, pywintypes.error, Exception):
             # ALWAYS! let user know what happened


Neizvestnyj:
> Please provide a way to reproduce this issue.

You need to create a folder that will be named using Cyrillic or Kanji and then open it in the file manager with the code below.

```python
from textwrap import dedent

from plyer import filechooser

from kivy.app import App
from kivy.lang import Builder


class FileChoose(Button):
    '''
    '''

    selection = ListProperty([])

    def choose(self):
        '''
        '''
        filechooser.choose_dir(on_selection=self.handle_selection)

        '''
        '''
        self.selection = selection

        '''
        '''


class ChooserApp(App):
    '''
    '''

    def build(self):
            <FileChoose>:
            BoxLayout:
                BoxLayout:
                    orientation: 'vertical'
                    TextInput:
                        id: result
                        text: ''
                        hint_text: 'selected path'
                    FileChoose:
                        size_hint_y: 0.1
                        on_release: self.choose()
                        text: 'Select a file'
        '''))


if __name__ == '__main__':
    ChooserApp().run()
```

PR #629, opened 2021-06-13
Add a `require_permissions` decorator to use on android methods

tshirtman:
adding the decorator with permissions on a method or function will automatically require the permissions when the method is called.

tested on the gps example, but as it worked, added on the other android implementations using the permissions requested in the buildozer.spec of their example.

### DIFF
--- a/examples/gps/main.py
+++ b/examples/gps/main.py
@@ -33,34 +33,6 @@ class GpsTest(App):
     gps_location = StringProperty()
     gps_status = StringProperty('Click Start to get GPS location updates')
 
-    def request_android_permissions(self):
-        """
-        Since API 23, Android requires permission to be requested at runtime.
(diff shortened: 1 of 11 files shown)


akshayaurora (approved):

Kulothungan16:
> Since API 23, Android requires permission to be requested at runtime.

what about this?

tshirtman:
> 
> 
> > Since API 23, Android requires permission to be requested at runtime.
> 
> what about this?

Hm, we should maybe check if the api level is < 23 and change the decorator to just return the initial method in that situation?
Are apps targeting API < 23 still accepted?

KINGMOHIT:
I am facing an issue microphone is not working with Android 10 or 10+ when api is set to 31 or more

PR #634, opened 2021-06-21
updated-device-name-implementation-for-backward-compatibility

ljnath:
Updated the previous `devicename` implementation for android.

Previous implementation was limited to API 17, updated the implementation so this this functionality works with API 1 as well.

### DIFF
--- a/plyer/platforms/android/devicename.py
+++ b/plyer/platforms/android/devicename.py
@@ -15,7 +15,7 @@ class AndroidDeviceName(DeviceName):
 
     def _get_device_name(self):
         """
-        Method to get the device name in an android environment.
+        Method to get the device name aka model in an android environment.
 
         Changed the implementation from 'android.provider.Settings.Global' to
         'android.os.Build' becuase 'android.provider.Settings.Global' was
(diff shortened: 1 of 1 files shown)


plyer/platforms/android/devicename.py
```diff
@@ -15,10 +14,16 @@ class AndroidDeviceName(DeviceName):
...
     def _get_device_name(self):
-        return Secure.getString(
-            activity.getContentResolver(),
-            Secure.DEVICE_NAME
-        )
+        """
+        Method to get the device name in an android environment.
+
+        Changed the implementation from 'android.provider.Settings.Global' to
+        'android.os.Build' becuase 'android.provider.Settings.Global' was
```
tshirtman, line 21:
```suggestion
        'android.os.Build' because 'android.provider.Settings.Global' was
```

tshirtman (commented):
Does that affect the result? Or should it be the same? if so, any reason for the two different apis that would be important to consider?

For regarding the comment, i think it’s more important to put such things in the git commit message than in the documentation, as the documentation is more about how to use it than how it changed over time.

plyer/platforms/android/devicename.py
ljnath: Thank you for your comment.
There is no change in the result, its the same.

tshirtman (approved):

PR #646, opened 2021-09-17
linux/storagepath: fixup a host of issues

rski:
The issues fixed are explained in the commits:

Author: Romanos Skiadas <rom.skiad@gmail.com>
Date:   Fri Sep 17 20:41:42 2021 +0300

    linux/storagepath: Revamp handling of user-dirs
    
    - dont assume that ~/.config/user-dirs.dirs exists
    - don't assume it does have contents or the relevant value
    - don't use exception handling as happy path control flow
    - when defaulting, use the path under HOME, not e.g. /Pictures

commit 99dabb2d62248fc3ea5705c2720abf71c9fc378b
Author: Romanos Skiadas <rom.skiad@gmail.com>
Date:   Fri Sep 17 20:31:08 2021 +0300

    linux/storagepath: Fixup DOCUMENTS to DOCUMENT
    
    this is the right name according to the XDG spec, and the key used in
    the default paths dict

This is a superset of https://github.com/kivy/plyer/pull/539. It both fixes more issues, and makes the code a bit simpler, as it removes the exception handling that is not really needed. The reaason for this PR is that the other PR seems to have stalled, and we would like to merge these fixes in order to package plyer for nixpkgs, as it's a dependency of a new version of mirage, which is already packaged.

### DIFF
--- a/plyer/platforms/linux/storagepath.py
+++ b/plyer/platforms/linux/storagepath.py
@@ -44,7 +44,7 @@ class LinuxStoragePath(StoragePath):
         return "/"
 
     def _get_documents_dir(self):
-        directory = self._get_from_user_dirs("DOCUMENT")
+        directory = self._get_from_user_dirs("DOCUMENTS")
         return directory.replace("$HOME", self._get_home_dir())
 
     def _get_downloads_dir(self):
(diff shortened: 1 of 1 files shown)


akshayaurora:
@rski 👍

PR #655, opened 2021-11-24
Bigger buffer, allows large selection

akshayaurora:
Fixes Crash due to low buffer for selecting large number of files > 100
Ref: https://mail.python.org/pipermail/python-win32/2018-June/014071.html


AndreMiras (approved):
LGTM

PR #665, opened 2022-02-11
Address LocationListener java method added with API31

HyTurtle:
For reference: [LocationListener](https://developer.android.com/reference/android/location/LocationListener#onLocationChanged(java.util.List%3Candroid.location.Location%3E))

Is there another way you'd handle this List/make use of the OnLocationChanged?

### DIFF
--- a/plyer/platforms/android/gps.py
+++ b/plyer/platforms/android/gps.py
@@ -31,14 +31,9 @@ class _LocationListener(PythonJavaClass):
 
     @java_method('(Ljava/util/List;)V', name='onLocationChanged')
     def onLocationChangedList(self, location_list):
+        ''' Called when location data is sent in a batch (API31) '''
         location = location_list.get(location_list.size() - 1)
-        self.root.on_location(
-            lat=location.getLatitude(),
-            lon=location.getLongitude(),
-            speed=location.getSpeed(),
-            bearing=location.getBearing(),
-            altitude=location.getAltitude(),
-            accuracy=location.getAccuracy())
+        self.onLocationChanged(location)
 
     @java_method('(Ljava/lang/String;)V')
     def onProviderEnabled(self, status):
--- a/plyer/platforms/android/gps.py
+++ b/plyer/platforms/android/gps.py
@@ -31,7 +31,7 @@ class _LocationListener(PythonJavaClass):
 
     @java_method('(Ljava/util/List;)V', name='onLocationChanged')
     def onLocationChangedList(self, location_list):
-        location = location_list[-1]
+        location = location_list.get(location_list.size() - 1)
         self.root.on_location(
             lat=location.getLatitude(),
             lon=location.getLongitude(),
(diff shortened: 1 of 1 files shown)


plyer/platforms/android/gps.py
```diff
@@ -29,6 +29,17 @@ def onLocationChanged(self, location):
             altitude=location.getAltitude(),
             accuracy=location.getAccuracy())
 
+    @java_method('(Ljava/util/List;)V', name='onLocationChanged')
```
Zen-CODE, line 32: Would be useful to have doc strings explaining the difference between `onLocationChanged` and `onLocationChangedList`? It's not obvious here...

Zen-CODE (changes requested):

tshirtman:
> For reference: [LocationListener](https://developer.android.com/reference/android/location/LocationListener#onLocationChanged(java.util.List%3Candroid.location.Location%3E))
> 
> Is there another way you'd handle this List/make use of the OnLocationChanged?

there might be apps that want to trace the path of the device as precisely as possible, and would rather have all the intermediate points? Maybe we could let the user decide which scenario they want to handle, either through a flag, or through a separate event.

tito:
For me it seems ok, better than plyer breaking on API 31. We are at least sync with previous implementation.

For the record, even targetting with SDK 30, this new method got triggered.

```
❯ aapt list -a build/bin/lillefives1942-0.9.0-armeabi-v7a-debug.apk|grep SdkVersion
    A: android:compileSdkVersion(0x01010572)=(type 0x10)0x1e
    A: android:compileSdkVersionCodename(0x01010573)="11" (Raw: "11")
      A: android:minSdkVersion(0x0101020c)=(type 0x10)0x15
      A: android:targetSdkVersion(0x01010270)=(type 0x10)0x1e
```

```
03-15 19:43:54.477 30534 30534 I python  : ===== Python/java method missing ======
03-15 19:43:54.477 30534 30534 I python  : Python class:<plyer.platforms.android.gps._LocationListener object at 0xaed0e868>
03-15 19:43:54.477 30534 30534 I python  : Java method name:onLocationChanged
03-15 19:43:54.477 30534 30534 I python  : Signature: (Ljava/util/List;)V
03-15 19:43:54.477 30534 30534 I python  : =======================================
03-15 19:43:54.477 30534 30534 I python  : 
03-15 19:43:54.480 30534 30534 I python  :  Traceback (most recent call last):
03-15 19:43:54.480 30534 30534 I python  :    File "jnius/jnius_proxy.pxi", line 50, in jnius.jnius.PythonJavaClass.invoke
03-15 19:43:54.480 30534 30534 I python  :    File "jnius/jnius_proxy.pxi", line 74, in jnius.jnius.PythonJavaClass._invoke
03-15 19:43:54.480 30534 30534 I python  :  NotImplementedError: The method ('onLocationChanged', ('V', ('Ljava/util/List;',))) is not implemented
```

I am accepting this.

ttsgehrke (commented):
I applied the changes to my Samsung Tab 8 and these changes worked well.  The code works well.

PR #666, opened 2022-02-19
Fix bug, when user canceled filechooser, `on_selection` does not dispatched

Neizvestnyj:
When clicking the close button, on OS X on_selection was not called
```py
from textwrap import dedent

from plyer import filechooser

from kivy.app import App
from kivy.lang import Builder


class FileChoose(Button):
    '''
    '''

    selection = ListProperty([])

    def choose(self):
        '''
        '''
        filechooser.open_file(on_selection=self.handle_selection)

        '''
        '''
        self.selection = selection

        '''
        '''
        print(self.selection)


class ChooserApp(App):
    '''
    '''

    def build(self):
            <FileChoose>:
            BoxLayout:
                BoxLayout:
                    orientation: 'vertical'
                    TextInput:
                        id: result
                        text: ''
                        hint_text: 'selected path'
                    FileChoose:
                        size_hint_y: 0.1
                        on_release: self.choose()
                        text: 'Select a file'
        ''')
                                   )


if __name__ == '__main__':
    ChooserApp().run()
```

https://user-images.githubusercontent.com/40869738/154804345-a0b47343-403c-4ee3-8aa5-9c909be5e19a.mp4

### DIFF
--- a/plyer/platforms/macosx/filechooser.py
+++ b/plyer/platforms/macosx/filechooser.py
@@ -98,17 +98,14 @@ class MacFileChooser:
             url = NSURL.fileURLWithPath_(self.path)
             panel.setDirectoryURL_(url)
 
-        selection = ['']
+        selection = None
 
         if panel.runModal():
             if self.mode == "save" or not self.multiple:
                 selection = [panel.filename().UTF8String()]
             else:
                 filename = panel.filenames()
-                selection = [
-                    filename.objectAtIndex_(x).UTF8String()
-                    for x in range(filename.count())
-                ]
+                selection = [filename.objectAtIndex_(x).UTF8String() for x in range(filename.count())]
 
         self._handle_selection(selection)
 
--- a/plyer/platforms/macosx/filechooser.py
+++ b/plyer/platforms/macosx/filechooser.py
@@ -98,18 +98,21 @@ class MacFileChooser:
             url = NSURL.fileURLWithPath_(self.path)
             panel.setDirectoryURL_(url)
 
+        selection = ['']
+
         if panel.runModal():
-            selection = None
             if self.mode == "save" or not self.multiple:
                 selection = [panel.filename().UTF8String()]
             else:
                 filename = panel.filenames()
                 selection = [
                     filename.objectAtIndex_(x).UTF8String()
-                    for x in range(filename.count())]
-            self._handle_selection(selection)
-            return selection
-        return None
+                    for x in range(filename.count())
+                ]
+
+        self._handle_selection(selection)
+
+        return selection
 
 
 class MacOSXFileChooser(FileChooser):
--- a/plyer/platforms/macosx/filechooser.py
+++ b/plyer/platforms/macosx/filechooser.py
@@ -105,7 +105,9 @@ class MacFileChooser:
                 selection = [panel.filename().UTF8String()]
             else:
                 filename = panel.filenames()
-                selection = [filename.objectAtIndex_(x).UTF8String() for x in range(filename.count())]
+                selection = [
+                    filename.objectAtIndex_(x).UTF8String()
+                    for x in range(filename.count())]
 
         self._handle_selection(selection)
 


tshirtman:
But, you didn't select anything, why should it be called?

Neizvestnyj:
> But, you didn't select anything, why should it be called?

Because on others platform (win/linux) its called and if the user clicked the close button there is no way to find out it

HyTurtle:
Does setting allownone and/or force_dispatch to True on you ListProperty not do the same? And without the issue created of defaulting to a True/Truthy value

Neizvestnyj:
> Does setting allownone and/or force_dispatch to True on you ListProperty not do the same? And without the issue created of defaulting to a True/Truthy value

No, `selection = ListProperty(['']) / selection = ListProperty([True])` does not work

tshirtman:
Doesn't seem to be the case on Linux? (i just tested and if i close the selection window, there is no change)
https://github.com/kivy/plyer/blob/master/plyer/platforms/linux/filechooser.py#L68
but on Windows indeed seems to be the case
https://github.com/kivy/plyer/blob/master/plyer/platforms/win/filechooser.py#L149
although it returns `[]` not `['']`.
on android, it seems it would dispatch [] as well.
https://github.com/kivy/plyer/blob/master/plyer/platforms/android/filechooser.py#L194
not sure on ios
https://github.com/kivy/plyer/blob/master/plyer/platforms/ios/filechooser.py#L77

Neizvestnyj:
> Doesn't seem to be the case on Linux? (i just tested and if i close the selection window, there is no change) https://github.com/kivy/plyer/blob/master/plyer/platforms/linux/filechooser.py#L68 but on Windows indeed seems to be the case https://github.com/kivy/plyer/blob/master/plyer/platforms/win/filechooser.py#L149 although it returns `[]` not `['']`. on android, it seems it would dispatch [] as well. https://github.com/kivy/plyer/blob/master/plyer/platforms/android/filechooser.py#L194 not sure on ios https://github.com/kivy/plyer/blob/master/plyer/platforms/ios/filechooser.py#L77

Yes, on different platform different return...

HyTurtle:
> > Does setting allownone and/or force_dispatch to True on you ListProperty not do the same? And without the issue created of defaulting to a True/Truthy value
> 
> No, `selection = ListProperty(['']) / selection = ListProperty([True])` does not work

that would be `selection = ListProperty([], allownone=True, force_dispatch=True)`

Neizvestnyj:
I can change in commit `selection = [''] to selection = []`, but in example code I need to set `selection = List Property([''])`

Neizvestnyj:
> selection = ListProperty([], allownone=True, force_dispatch=True)

does not work

HyTurtle:
Could adjust your commit so it passes None to _handle_selection - that shouldn't _break_ how it currently operates.

Neizvestnyj:
> 

```py
selection = None

if panel.runModal():
    if self.mode == "save" or not self.multiple:
        selection = [panel.filename().UTF8String()]
    else:
        selection = [i.UTF8String() for i in panel.filenames()]

self._handle_selection(selection)

return selection
```

```py
# in app
selection = ListProperty([], allownone=True)
```
so?

HyTurtle:
Seems harmless to me, does that selection trigger / not trigger or error without the allownone. and _handle_selection is a dummy method so in app you can redefine that if need be.

Neizvestnyj:
> Seems harmless to me, does that selection trigger / not trigger or error without the allownone. and _handle_selection is a dummy method so in app you can redefine that if need be.

Thanks, done

HyTurtle:
Just tried to read the relevant docs here, mind testing `print(panel.runModal())` with a selection made and an exit; won't matter too much just curious if the exit/cancel is just 0.

Neizvestnyj:
> Just tried to read the relevant docs here, mind testing `print(panel.runModal())` with a selection made and an exit; won't matter too much just curious if the exit/cancel is just 0.

hmm, yes, I see

tshirtman (approved):
LGTM

Neizvestnyj:
> LGTM
> 
> please fix the style issue though, (line > 80 breaks pep8)

Ah, yes, I see, fixed it)

PR #667, opened 2022-02-27
Added the ability to track the closure of the file manager without selecting content

Neizvestnyj:
When clicking the close button, on *Linux* `on_selection` was not called. Same - #666 
```py
from textwrap import dedent

from plyer import filechooser

from kivy.app import App
from kivy.lang import Builder


class FileChoose(Button):
    '''
    '''

    selection = ListProperty([], allownone=True)

    def choose(self):
        '''
        '''
        filechooser.open_file(on_selection=self.handle_selection)

        '''
        '''
        self.selection = selection

        '''
        '''
        print(self.selection)


class ChooserApp(App):
    '''
    '''

    def build(self):
            <FileChoose>:
            BoxLayout:
                BoxLayout:
                    orientation: 'vertical'
                    TextInput:
                        id: result
                        text: ''
                        hint_text: 'selected path'
                    FileChoose:
                        size_hint_y: 0.1
                        on_release: self.choose()
                        text: 'Select a file'
        ''')
                                   )


if __name__ == '__main__':
    ChooserApp().run()

```

### DIFF
--- a/plyer/platforms/linux/filechooser.py
+++ b/plyer/platforms/linux/filechooser.py
@@ -64,14 +64,15 @@ class SubprocessFileChooser:
             if ret is not None:
                 if ret == self.successretcode:
                     out = self._process.communicate()[0].strip().decode('utf8')
-                    self.selection = self._split_output(out)
-                    self._handle_selection(self.selection)
-                    return self.selection
+                    return self._set_and_return_selection(self._split_output(out))
                 else:
-                    self.selection = None
-                    self._handle_selection(self.selection)
-                    return self.selection
+                    return self._set_and_return_selection(None)
             time.sleep(0.1)
+
+    def _set_and_return_selection(self, value):
+        self.selection = value
+        self._handle_selection(value)
+        return value
 
     def _split_output(self, out):
         '''This methods receives the output of the back-end and turns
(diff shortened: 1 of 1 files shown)


plyer/platforms/linux/filechooser.py
```diff
@@ -68,7 +68,9 @@ def _run_command(self, cmd):
                     self._handle_selection(self.selection)
                     return self.selection
                 else:
-                    return None
+                    self.selection = None
+                    self._handle_selection(self.selection)
+                    return self.selection
```
tshirtman, line 73: seems like it would be worth extracting that duplicate logic to a helper method. e.g:
```py
def set_and_return_selection(self, value):
    self.selection = value
    self._handle_selection(value)
    return value
```

and use it
```py
    if ret is not None:
        ...
        return self.set_and_return_selection(self._split_output(out)
    else:
        return self.set_and_return_selection(None)
```

Neizvestnyj: Done

https://user-images.githubusercontent.com/40869738/160231054-b827071b-af6b-4104-9add-98c1782b13ea.mp4

tshirtman (approved):

PR #670, opened 2022-03-12
Use sys.getandroidapilevel for more robust Android detection

rdb:
Currently plyer uses the presence of the `ANDROID_ARGUMENT` environment variable to detect Android.  As of Python 3.7, there is a `sys.getandroidapilevel` that can be checked instead, which is more reliable since it does not require on a particular implementation detail of p4a.


akshayaurora:
@rdb thanks!

PR #672, opened 2022-03-26
android 11+ compatibility Documents folder

moonpyx:
for android 11+ it is not possible to capture the real path of the files inside the documents folder, this implementation solves this scenario.


Zen-CODE (approved):
Nice. Thanks!

PR #692, opened 2022-07-10
Fixes a failing test for notification

misl6:
- `hints` is a `dict`, but during the test a `list` was expected.


misl6:
See: #639

PR #694, opened 2022-07-16
Add missing platforms in audio

rshah713:
Add missing platforms in audio docstring

### DIFF
--- a/plyer/facades/audio.py
+++ b/plyer/facades/audio.py
@@ -41,7 +41,7 @@ To play recording::
 
 Supported Platforms
 -------------------
-Android, Windows, OS X
+Android, Windows, macOS
 
 '''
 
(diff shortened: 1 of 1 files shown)


plyer/facades/audio.py
```diff
@@ -41,7 +41,7 @@
 
 Supported Platforms
 -------------------
-Android
+Android, Windows, OS X
```
misl6, line 44:
```suggestion
Android, Windows, macOS
```

misl6 (commented):
Left a minor suggestion :)

plyer/facades/audio.py
rshah713: Thanks for the review @misl6! I used `devicename`as reference but this totally works

misl6 (approved):
LGTM. Thank you!

Issue #698, opened 2022-07-28
unformatted content on pypi page


totallynotadi:
I've been concerned about the pypi page for plyer. The markdown is not rendered properly and the license shows up at the top unlike in that in the readme. Everything renderes properly on the github readme but not on the pypi page. there's also no way to read a short, well formatted intorduction for people discovering the pypi page first since it doesnt lead to the github repo. A shorter, more comprehensive intro is better to be on the front pages imo as opposed to the readthedocs, but pypi messes up formatting.

![this](https://pypi.org/project/plyer/) is how the page looks. There's a couple of issues with it - 
* license shows up unnecessarily.
* markdown is completely disregarded altogether.

it would be nice to see a sweet pypi page for plyer but i have no idea what's causing these issues
my best guess is regarding the `package_data` parameter from `setup.py` combined with the `manifest.in` file causing pypi to render things this way.

the following are the contents from `setup.py`
```
setup(
    name='plyer',
    version=plyer.__version__,
    description='Platform-independent wrapper for platform-dependent APIs',
    long_description=README + u"\n\n" + CHANGELOG + u"\n\n",
    long_description_content_type='text/markdown',
    author='Kivy team',
    author_email='mat@kivy.org',
    url='https://plyer.readthedocs.org/en/latest/',
    packages=PACKAGES,
    package_data={'': ['LICENSE', 'README.md']},
    package_dir={'plyer': 'plyer'},
    include_package_data=True,
    license=open(join(CURDIR, 'LICENSE')).read(),
    zip_safe=False,
    classifiers=[
        'Programming Language :: Python :: 3.5',
        'Programming Language :: Python :: 3.6',
        'Programming Language :: Python :: 3.7',
        'Programming Language :: Python :: 3.8',
    ]
)
```

and `manifest.in` just contains

```
include *README.rst *LICENSE
include *CHANGELOG.md
```

appreciate help for ideas to resolve this issue, thanks

Julian-O:
Is fixed now, perhaps by the PR associated with #701.

Closing as complete.

PR #701, opened 2022-08-09
Keyword should only hold name of license

rshah713:
In accordance with pypa/twine#425, `license` keyword [should not](https://github.com/pypa/twine/issues/425#issuecomment-466436196) hold a multi-line file but rather simply be the name for the license. 

Could be potential fix for kivy/plyer#698 where raw Markdown is being displayed on PyPI. I am definitely interested in working for a fix for kivy/plyer#698 but would appreciate some guidance in where to start as I am more of a beginner.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ setup(
     package_data={'': ['LICENSE', 'README.md']},
     package_dir={'plyer': 'plyer'},
     include_package_data=True,
-    license=open(join(CURDIR, 'LICENSE')).read(),
+    license='LICENSE',
     zip_safe=False,
     classifiers=[
         'Development Status :: 4 - Beta',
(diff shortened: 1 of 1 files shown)


tshirtman:
Good call, but the name of the license is "MIT" not "LICENSE", but if you change that, i think the PR should be accepted.

setup.py
```diff
@@ -66,7 +66,7 @@
     package_data={'': ['LICENSE', 'README.md']},
     package_dir={'plyer': 'plyer'},
     include_package_data=True,
-    license=open(join(CURDIR, 'LICENSE')).read(),
+    license='LICENSE',
```
tshirtman, line 69:
```suggestion
    license='MIT',
```

rshah713:
Whoops, that makes more sense. Thanks for the review @tshirtman!

tshirtman:
Thank you :)

PR #704, opened 2022-08-29
Document supported platforms for humidity

rshah713:


plyer/facades/humidity.py
```diff
@@ -4,6 +4,10 @@ class Humidity:
        With method `enable` you can turn on Humidity sensor and
        'disable' method stops the sensor.
        Use property `tell` to get humidity value.
+       
+       Supported Platforms
+       -------------------
+       Linux
```
tshirtman, line 10: Didn’t you mean android? i only see this one https://github.com/kivy/plyer/blob/master/plyer/platforms/android/humidity.py

rshah713: Yep, that's exactly what I meant. Worked on this way too late at night...

tshirtman (approved):

Issue #731, opened 2022-12-31
Feature request: navigation intent


antorix:
Hi, this is an awesome module. But I would greatly appreciate if you add the navigation intent call. I.e., the way to launch a navigation route in the navigation app chosen by the user among those installed on the device.

PR #732, opened 2023-01-02
Implement Maps facade

rshah713:
Provided is the general Maps facade structure.

I have chosen to create each platform extension in a separate PR as this is my first time creating a facade from scratch on this project. This will separate issues from implementing the facade vs implementing it on each platform.


rshah713:
Also moves to fix #731.

plyer/facades/maps.py
```diff
@@ -0,0 +1,83 @@
...
+The :class:`Maps` creates a client for accessing the default Maps API.
+
+Holds features such as opening a location by
+address & latitude/longitude, create queries, or find directions between
+two points
+
+Simple Examples
+---------------
+
+Perform a search::
```
misl6, line 13: Maybe we should specify that the search results are not returned, but instead the OS default map app gets opened, and will show the search results?

plyer/facades/maps.py
```diff
@@ -0,0 +1,83 @@
...
+
+    >>> from plyer import maps
+    >>> maps.search('Mexican Restaurant')
+    >>> maps.search('Taco Bell', latitude=38.5810606, longitude=-121.493895)
```
misl6, line 17: Better to avoid using a real company trade name here in our docs.

misl6 (changes requested):
Hi!
Left some minor suggestions, let me know what do you think about it.

plyer/facades/maps.py
rshah713: Great point, changed.

misl6 (commented):
The facade looks good (maybe we can add some tests), but I think we should keep it on stand-by until there's an implementation for at least a platform.

Thank you!

rshah713:
I was planning to open a separate PR for the platforms but I can add one to this PR to get it merged :)

plyer/platforms/macosx/maps.py
```diff
@@ -0,0 +1,90 @@
...
+        query = quote_plus(query, safe=',')
+        maps_address = 'http://maps.apple.com/?q=' + query
+
+        if latitude is not None and longitude is not None:
+            maps_address += '&sll={},{}'.format(latitude, longitude)
+
+        process = Popen(
+            ['open', '-a', 'Maps', maps_address],
+            stdout=PIPE, stderr=PIPE)
+        stdout, stderr = process.communicate()
```
misl6, line 68: What about:
```python
import webbrowser
webbrowser.open("http://maps.apple.com/?q=Tour Eiffel")
```

That will also be likely (did not tested yet, but should work) portable to iOS.

plyer/platforms/macosx/maps.py
```diff
@@ -0,0 +1,90 @@
+'''
+Module of MacOSX API for plyer.maps.
```
misl6, line 2:
```suggestion
Module of macOS API for plyer.maps.
```

misl6 (commented):
Left some comments regarding it.
Let me know what you think about it.

Maybe: We can also add a minimal set of tests?

plyer/platforms/macosx/maps.py
rshah713: I did consider this but by this route there is a dialogue that pops up every time asking "Would you like to open this in Maps" while my current method specifies it via the Maps app and it opens w/o any dialogue. 

All in all, minor but I personally prefer this as it's cleaner and more direct. Let me know your thoughts on this @misl6

plyer/platforms/macosx/maps.py
rshah713: done

rshah713:
@misl6 I left a comment on using `webbrowser` to open the address. 

At the moment, I don't have the bandwidth or time to learn & write tests for this unfortunately. However, I can look into creating a quick kivy app example in a later PR for future reference.

rshah713:
@misl6 Friendly ping :)

antorix:
Is it too early to try this update? I pip-installed:

`https://github.com/kivy/plyer/archive/refs/heads/master.zip`

But when I do

`from plyer import maps`

I get

`cannot import name 'maps'`

rshah713:
@antorix This PR has 
A). Not been merged
B). Not been added to the new PyPi release. 

To use the changes in this PR, you can either wait for it to be merged at which point you can clone the master branch and use it. Or you can directly add the changed files in this PR yourself despite it not being merged.

If you add the changes in this PR to your own project and it works, please report back in this thread as it will likely get this PR merged faster :)

plyer/platforms/macosx/maps.py
misl6: > I did consider this but by this route there is a dialogue that pops up every time asking "Would you like to open this in Maps" while my current method specifies it via the Maps app and it opens w/o any dialogue.

Not happening on my side, but it may be a specific configuration / "Don't ask anymore" flag.

`open -a Maps` is not documented anywhere, and this is why I'm worried about future potential headaches.
But, everything can change (even the link structure), so feel free to use your implementation, if you're fine with it 😀

misl6 (approved):
LGTM. Thank you!

antorix:
Sorry, but I still can't understand how to use it. I'm doing this:

```
from plyer import maps
maps.route('Here', address) # where address is a string containing actual street and number
```

The program crashes at this point. What am I doing wrong? I include plyer in buildozer using `requirements = https://github.com/kivy/plyer/archive/refs/heads/master.zip`

rshah713:
Android platform is currently not supported. MacOS is fully supported, and I have #750 pending review for iOS. Feel free to open a PR for android implementation as the facade is already there.

Issue #735, opened 2023-01-12
FileChooser raises NotImplementedError/ModuleNotFoundError without pywin32 on windows 10


nWestie:
When using plyer FileChooser without kivy, it raises an error when it cannot find the win32 module. After installing pywin32, it works fine. Could pywin32 be added as a dependancy of plyer, to reduce confusion? Kivy installs it automatically on windows.
See also: #602
the full error/traceback:

> Traceback (most recent call last):
  File "C:\Users\<USER>\scripts\.venv\lib\site-packages\plyer\utils.py", line 96, in _ensure_obj
    mod = __import__(module, fromlist='.')
    from win32com.shell.shell import (
ModuleNotFoundError: No module named 'win32com'
Traceback (most recent call last):
  File "c:\Users\WESTNB21\Scripts\linecounter.py", line 3, in <module>
    fileName = filechooser.open_file(
    return self._file_selection_dialog(mode="open", *args, **kwargs)
  File "C:\Users\<USER>\scripts\.venv\lib\site-packages\plyer\facades\filechooser.py", line 74, in _file_selection_dialog
    raise NotImplementedError()
NotImplementedError

I am on windows 10, with python 3.10

misl6:
Hi @nWestie 

> Could pywin32 be added as a dependancy of plyer, to reduce confusion? Kivy installs it automatically on windows.

Yes, do you want to take care of the PR and testing? (I do not have a Windows setup available ATM)

nWestie:
sure, I think I can figure out how to do that, I'll give it a shot this week sometime

PR #737, opened 2023-01-17
Add pywin32 as optional dependancy

nWestie:
Fix for #735 

add pywin32 dependency when installing with the [win] option.
Alternative, it could be added as a platform specific dependency for windows(added, but commented out)

I'm not sure which is the most ideal behavior, thoughts?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ try:
                 'ios': ['pyobjus'],
                 'macosx': ['pyobjus'],
                 'android': ['pyjnius'],
+                'win': ['pywin32'],
                 'dev': ['mock', 'flake8']
             }
         }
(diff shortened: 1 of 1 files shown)


misl6:
For now, in order to target issue #735 adding it as an optional dependency is fine (so please remove the additional comment).

In near future (and if you want to, feel free to contribute), I think is better to install platform-specific dependencies by default, in order to be more developer friendly. (Unless the footprint of the platform-specific dependency is huge)

Also, we should consider switching to https://setuptools.pypa.io/en/latest/userguide/declarative_config.html

nWestie:
Great.
I'd agree that it probably makes sense to add platform dependencies automatically, my only question is how that looks when developing for android/iOS; I assume those would stay as optional dependencies, and the dependencies for mac/win/linux would be installed anyways? 

As far as updating, would it make more switch to the `pyproject.toml` standard instead? 
  https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
My impression is that it is the most up to date format, so it could make sense to switch to that if there is no specific reason not to

PR #755, opened 2023-03-29
Fixes error when selecting a file from the `Documents` option in newer Android APIs

DexerBR:
When selecting a file in the `Documents` option in the navigation panel, the `uri_authority` returned is `'com.android.providers.media.documents'`. However the selection will fail because the `_handle_media_documents` method only handles `audio`, `video` or `image` and ignores any other type of file. Finally, `None` will be returned as the file path.

### Tested on:

`python-for-android` version: `2023.2.10`
`buildozer` version: `1.5.0`

**Android API levels:**
- `27`
- `28`
- `31`


akshayaurora (approved):

akshayaurora:
@DexerBR thanks 👍

Issue #702, opened 2022-08-23
plyer notification android api 31 error


wlyapps:
android api 31 error!

plyer notification has the following error on android api 31:

JVM exception occurred: org.testeapps.controllapp:
Targeting S+ (version 31 and above) requires that one
of FLAG_IMMUTABLE or FLAG_MUTABLE be specified
when creating a Pendingintent. Strongly consider using
FLAG_IMMUTABLE, only use FLAG_MUTABLE if some

functionality depends on the Pendingintent being
mutable, e.g. if it needs to be used with inline replies or
bubbles. java.lang.IllegalArgumentException

LazzaAU:
i have the same issue and guessing this is the cause of not getting notification when i run the app on android. Was there any thoughts on where i'd need to add the flag's ? or is that something that needs done internally of plyer ?

LazzaAU:
Is there any plans to fixing this ? I can't set my API level to anything above 30 or this error kicks in. Google tells me i cant upload a aab file to the play store if the API is under 31. So i'm kinda stuck on the fence.

Other than setting up push notifications any suggestions to work around this ?

Zahrafathi1234:
I have the same issue. Did you find a solution for it? I asked my question here.  Can the problem be solved?
https://stackoverflow.com/questions/75532341/flag-immutable-or-flag-mutable-in-notification-kivy-for-android
I am a new Kivy user and I don't know where the problem is.

SomberNight:
feel free to test with https://github.com/kivy/plyer/pull/756 - that works for me

seba-plpl:
hi, i tried to test it using https://github.com/HyTurtle/plyer/archive/master.zip in requirements but no luck, i have this issue: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.

please tell me, how to test it?

SomberNight:
That fork looks too old. Try using kivy/plyer.

rdb:
It looks like there has not been a Plyer release since this issue was fixed in 2023, is it possible to push a new release to PyPI? This is currently a blocker for the ability to use notifications on Android since it's no longer possible to upload new apps targeting SDK level 30.

PR #756, opened 2023-04-06
android: fix notification.notify() for sdk>=31, by setting new req flag

SomberNight:
fixes https://github.com/kivy/plyer/issues/702

see https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability
> Pending intents mutability
>   If your app targets Android 12, you must specify the mutability of
>   each PendingIntent object that your app creates. This additional
>   requirement improves your app's security.

see https://stackoverflow.com/q/69088578

Without this, I get the following exception: `JavaException('JVM exception occurred: org.electrum.testnet.electrum: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.\nStrongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles. java.lang.IllegalArgumentException')`

As in code comment, [`PendingIntent.FLAG_IMMUTABLE`](https://developer.android.com/reference/android/app/PendingIntent#FLAG_IMMUTABLE) was introduced in sdk 23.
Prior to sdk 31, if `FLAG_IMMUTABLE` is not set, the intent is implicitly mutable. Starting with sdk 31, either `PendingIntent.FLAG_IMMUTABLE` or `PendingIntent.FLAG_MUTABLE` needs to be explicitly specified.
AFAICT, we can just always set `FLAG_IMMUTABLE` (if sdk>=23), which is what this PR does. However, I have only tested on Android 13 (api level 33).

### DIFF
--- a/plyer/platforms/android/notification.py
+++ b/plyer/platforms/android/notification.py
@@ -148,10 +148,15 @@ class AndroidNotification(Notification):
         notification_intent.setFlags(Intent.FLAG_ACTIVITY_SINGLE_TOP)
         notification_intent.setAction(Intent.ACTION_MAIN)
         notification_intent.addCategory(Intent.CATEGORY_LAUNCHER)
+        if SDK_INT >= 23:
+            # FLAG_IMMUTABLE added in SDK 23, required since SDK 31:
+            pending_flags = PendingIntent.FLAG_IMMUTABLE
+        else:
+            pending_flags = 0
 
         # get our application Activity
         pending_intent = PendingIntent.getActivity(
-            app_context, 0, notification_intent, 0
+            app_context, 0, notification_intent, pending_flags
         )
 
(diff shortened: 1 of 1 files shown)


akshayaurora (approved):

akshayaurora:
👍

seba-plpl:
hi, i tried to test it using https://github.com/HyTurtle/plyer/archive/master.zip in requirements but no luck, i have this issue: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.

please tell me, how to test it?

PR #781, opened 2023-11-01
Fix `_get_sdcard_dir` for Android API < 24

DexerBR:
`getStorageVolumes` is available starting from [API 24](https://developer.android.com/reference/android/os/storage/StorageManager#getStorageVolumes()), so the code needs to be fixed to work on lower APIs.


misl6:
The PEP8 Check is complaining, but not your fault.

misl6 (approved):
LGTM. Thank you!

PR #782, opened 2023-11-01
Switch to MaxFile arg to increase Windows filechooser buffer

Xnot:
In d87c446, a workaround to long filenames crashing the Windows filechooser was added by appending `\x00` to the `File` arg. This works because [pywin32 takes the max of `MaxFile` and the length of `File`](https://github.com/mhammond/pywin32/blob/4025d7680552beb200688911e3ec87e11acf9365/win32/src/win32gui.i#L6544) as the buffer size. However it would be more appropriate to directly use the `MaxFile` arg, as that's its intended purpose. Additionally, this will allow users to pass the `buffer_size` parameter if they are using really long file names and wish to override the buffer size.


akshayaurora:
@Xnot thanks,  sry for the late reply on this.

Issue #778, opened 2023-10-27
Remove references to distutils


Julian-O:
[setup.py](https://github.com/kivy/plyer/blob/master/setup.py) attempts to import `setuptools`, and if it fails imports from `distutils`.

`distutils` is deprecated and no longer present in Python 3.12.

Assume setuptools is available, and remove the `distutils` reference.

PR #784, opened 2023-11-15
Remove references to distutils

caseycronyn:
#778

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,25 +14,21 @@ For Android install additional dependency PyJNIus::
 
 from os.path import dirname, join
 import plyer
+from setuptools import setup
 import io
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -14,25 +14,21 @@
 
 from os.path import dirname, join
 import plyer
+from setuptools import setup
 import io
 
 EXTRA_OPTIONS = {}
```
misl6, line 20: I guess we do not need this one anymore 😀

And, since we now only use `setuptools`, how about moving the whole extra options into `setup(` ?

misl6 (changes requested):
Hi @caseycronyn !

I've left a comment, let me know what do you think about it 😀

PR #785, opened 2023-11-21
Keystore now encrypts data before storing it, also added an example

t0theRANCH:


misl6 (changes requested):
Hi @t0theRANCH !

Thank you for taking care of this feature, can you please add some more context by editing the PR description?

Also, I see you're including some `dylibs`, it's preferable to avoid including pre-built binaries, but instead have a build script (that can run in the CI), can you take care of it?

t0theRANCH:
I need to use the dylib to import a framework and add some functions to a custom class that I load in the iOS key store class. Can the CI accomplish this? I'm not really sure what it is or how it works.

As for the PR description, do you just want some more details as to how exactly the iOS Keyring and Android Keystore encrypt and save information?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

misl6:
> As for the PR description, do you just want some more details as to how exactly the iOS Keyring and Android Keystore encrypt and save information?

Yes, just some details + links to resources (if applicable), just to give some context to someone else in future.

> I need to use the dylib to import a framework and add some functions to a custom class that I load in the iOS key store class. Can the CI accomplish this? I'm not really sure what it is or how it works.

I see, and unfortunately we can't ship any `.dylib` to iOS App store, unless the `.dylib` is placed into a (Signed) umbrella framework.
The suggestion (the most immediate and the faster way ATM) is to add the needed code in https://github.com/kivy/kivy-ios/tree/master/kivy_ios/recipes/ios/src , so it's available in plyer later when needed.

PR #790, opened 2023-12-11
Choose a reasonable and valid number for the intent code and use the uri variable

matan-h:
This PR make the `Android file chooser` use a reasonable and valid number in the intent code.

This PR contains another minor changes: in the `_resolve_uri` function, if the `uri` came from the `com.android.providers.media.documents` provider, it defines the `uri` variable, but don't actually use it, and return `path` which is still `None`, so this minor change make this method return `path or uri`.

### DIFF
--- a/plyer/platforms/android/filechooser.py
+++ b/plyer/platforms/android/filechooser.py
@@ -107,8 +107,8 @@ class AndroidFileChooser(FileChooser):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.select_code = randint(123456, 654321)
-        self.save_code = randint(123456, 654321)
+        self.select_code = randint(100, 65536) # 65536 is the maximum. 100 to avoid conflicts.
+        self.save_code = randint(100, 65536)
         self.selection = None
 
         # bind a function for a response from filechooser activity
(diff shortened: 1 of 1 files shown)


Julian-O:
I am not 100% sure I understand the context on this code, so let me explain what I think:
* The save_code and result_code are meant to be unique identifiers, that show that the callback is related to this particular call. 
* The Pythonic way would be to use an `object()` call to guarantee uniqueness within the Python application. I *believe* the other callers are not coming from Python, so this technique isn't guaranteed.
* Using a `uuid` would be another standard approach to avoid clashes. I am not clear whether there is a hardware limit that makes this impossible.
* The old code just generates two random 6 digit numbers and just hopes there isn't a clash.
* The old code just generates two random 5 digit numbers and just hopes there isn't a clash.

My comments:
* 5 digits doesn't feel rare enough. There will be clashes in the field.
* Your edit says 65536 is the maximum, but doesn't explain what is imposing this maximum. Given it is 2^16, it sounds like it must be an 16-bit OS limitation. If so, you should explain (link?) to why it is so limited. If not, how about 2^32 instead?
* save_code and result_code can randomly be generated as the same number, which will cause problems. How about make the top end of the range of the first one one smaller, and then define the second one as the first one + 1? That way they never clash.

matan-h:
Thank you, @Julian-O. I've added an explanation to the code and edited the description of this pull request to make it cleaner.

I've also removed the other minor change as that would probably require a design change.

Julian-O:
Thanks. I understand now.

[Just to be clear, I am not an official reviewer with merge rights. We'll need to wait for one of them.]

PR #795, opened 2023-12-29
Normalize path when passing args to windows filechooser

Xnot:
This changes the Win32FileChooser to apply `os.path.normpath` to the path parameter before passing it on to pywin32. This allows the filechooser's path parameter to work properly without the end-user having to normalize the path first.


akshayaurora (approved):

akshayaurora:
@Xnot thanks

PR #801, opened 2024-01-24
Exact match lookup for MacOS battery status

### LINKED ISSUE #799: Mac OS battery status
When I try out the battery.status on Mac OS it seems like the CurrentCapacity is the actually current battery percentage already, the calculation for percentage is not needed? 

Also, not sure if anyone else on Mac OS have the same issue

```
    max_capacity = float(line.rpartition('=')[-1].strip())
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: '80}'
```

There is an extra } at the end, my fix if to add a string slice to drop the }

`max_capacity = float(line.rpartition('=')[-1].strip()[:-1])`

To summarize this is the changes I made to have an output for battery status _{'isCharging': True, 'percentage': 98.0}_
```
for line in output.decode('utf-8').splitlines():
    if 'IsCharging' in line:
        is_charging = line.rpartition('=')[-1].strip()
    if 'MaxCapacity' in line:
        max_capacity = float(line.rpartition('=')[-1].strip()[:-1])
    # if 'CurrentCapacity' in line:
    if 'CurrentCapacity' == line.rpartition('=')[0].strip()[1:-1]:
        current_capacity = float(line.rpartition('=')[-1].strip())

if is_charging:
    status['isCharging'] = is_charging == "Yes"

if current_capacity and max_capacity:
```
    # status['percentage'] = 100.0 * current_capacity / max_capacity
    status['percentage'] = current_capacity

Ocean48:
Fix #799
Use exact string math instead of partial match to look for battery status. This will prevent fields like AppleRawMaxCapacity and AppleRawCurrentCapacity from being read as MaxCapacity and CurrentCapacity.

### DIFF
--- a/plyer/platforms/macosx/battery.py
+++ b/plyer/platforms/macosx/battery.py
@@ -32,11 +32,11 @@ class OSXBattery(Battery):
 
         is_charging = max_capacity = current_capacity = None
         for line in output.decode('utf-8').splitlines():
-            if 'IsCharging' in line:
+            if 'IsCharging' == line.rpartition('=')[0].strip()[1:-1]:
                 is_charging = line.rpartition('=')[-1].strip()
-            if 'MaxCapacity' in line:
+            if 'MaxCapacity' == line.rpartition('=')[0].strip()[1:-1]:
                 max_capacity = float(line.rpartition('=')[-1].strip())
-            if 'CurrentCapacity' in line:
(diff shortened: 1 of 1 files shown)


Julian-O (approved):
I am not an official core team member; I can't approve merges.

However, I have been following the diagnosis of this bug, and approve the changes.

PR #802, opened 2024-02-04
Secure Password Encryption and Storage for iOS and Android 

t0theRANCH:
## Secure Password Encryption and Storage for iOS and Android with Plyer

This Pull Request introduces secure password storage capabilities to the Plyer library, leveraging the native security features of iOS Keyring and Android Keystore systems. It allows for the encryption and secure storage of passwords and sensitive information, simplifying the process for developers to manage secure data on mobile devices.

### iOS Implementation

Utilizes the iOS Keychain Services for encrypted storage, providing a secure enclave for sensitive information. This ensures data protection even in the event of device compromise, leveraging the platform's encryption capabilities. In order to get this to work, I had to write an Objective-C class that contains the functions to interact with the keychain. They were a part of a framework, and not a class, and are thus, unable to be accessed with pyobjus normally. I compiled it into a .framework to comply with Apple's App Store guidelines. This will give you a bridge from Objective-C to Python that will allow you to interact with the keychain and automatically encrypt the data before storing it securely according to iOS best practices.

It comes with three functions:

    - (BOOL)saveWithService:(NSString *)service account:(NSString *)account value:(NSString *)value;
    - (NSString *)retrieveWithService:(NSString *)service account:(NSString *)account;
    - (BOOL)deleteWithService:(NSString *)service account:(NSString *)account;

For more information, see:
https://developer.apple.com/documentation/security/keychain_services/

### Android Implementation

Integrates with the Android Keystore system to securely manage and store cryptographic keys, using these keys to encrypt data. This approach keeps cryptographic keys safe and ensures that passwords are securely encrypted. The primary function of the Android Keystore system is to secure cryptographic keys. The keys stored in the Keystore are protected from extraction using the device, even if the device itself is compromised. This is achieved by isolating the keys in a secure hardware component (the Trusted Execution Environment, TEE, or on devices with a Secure Element, SE) or by using software-based protections if hardware support is not available.

The Keystore system allows apps to perform cryptographic operations such as encryption, decryption, signing, and verification using the stored keys without exposing the key material to the app. This means the cryptographic operations are performed by the Keystore, and at no point does the app need to handle raw key material, enhancing security. Key material never leaves the secure hardware and is not accessible to the app.

Developers can generate new keys directly in the Keystore, import existing keys into the Keystore, and manage keys (such as setting validity periods or usage restrictions). The Keystore supports various key types, including RSA, EC, and AES, although this implementation just uses AES.

For more information, see:
https://developer.android.com/privacy-and-security/keystore
https://developer.android.com/reference/android/security/keystore/KeyProperties
https://developer.android.com/reference/javax/crypto/KeyGenerator
https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec
https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder
https://developer.android.com/reference/javax/crypto/spec/GCMParameterSpec

### Usage

This feature provides a simple interface for encrypting and storing passwords:

```python
from plyer import keystore

"""
Encrypt and save a password
'service_identifier' is pretty much arbitrary, you can change it to whatever you want
change 'key' to whatever data you want to save (i.e. password, token, etc.)
'value' is the string you want to encrypt
"""
keystore.set_key('service_identifier', 'key', 'value')

# Retrieve and decrypt a password
password = keystore.get_key('service_identifier', 'key')


misl6:
Nice feature!

However, adding pre-compiled binaries to the repo should be avoided unless that's the only option.
Can you please provide the `xcodeproj` for the supplied framework instead?

t0theRANCH:
Sure can, do you want the entire project folder? Where do you want it?

Is there anything else you need me to do in regards to implementation? Sorry I am not that familiar with how plyer works. After figuring all of this out I thought it would be nice to share it.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

herbe13:
something new about this implementation? like when will be done?

Issue #668, opened 2022-03-07
How to turn on a flashlight on a phone with Android 9?


Andrukva:
Please tell me how to turn on the flashlight on a phone with Android 9 using plyer and kivy/switch or kivy/button. Those examples that exist do not work. The application starts, but as soon as I press the button to turn on the flashlight, the application closes. Please show me how to do it correctly. Thank you in advance!

Andrukva:
I requested permission when assembling in a bulldozer, as shown in the example.

Andrukva:
I requested permission for both the camera and the flashlight.

HyTurtle:
Just checked and `flash.on()` and `flash.off()` runs fine here on Android 11. Did you request camera at runtime ie within the app.
If not you can do so manually with your current app in `Settings > Apps`

Andrukva:
Could you show me what code you tried?

Andrukva:
Strangely, my application is not in the list of applications that are allowed or not allowed to use the camera. Although in the specification of the bulldozer I specified both the camera and the flash

oukiar:
I think this issue was solved by this pull request #716 

Tested under android 5 and android 9.

If you want to test change you requirements in your buildozer.spec
`requirements = kivy, plyer==https://github.com/oukiar/plyer`

and do `buildozer android clean` 

best regards

oukiar:
Some updates, today i was able to test in android 9 and my surprise is that using the legacy camera api not work.

For get working the flashlight was added code for use the CameraManager (newer api) now it is working and tested on android 9 device, maybe will work on other android versions, tests are necesary.

PR #806, opened 2024-04-02
Fix Flash switch issue on newer API #668

MeBlackLi:


plyer/platforms/android/flash.py
```diff
@@ -7,13 +7,20 @@
...
-SurfaceTexture = autoclass("android.graphics.SurfaceTexture")
-PackageManager = autoclass('android.content.pm.PackageManager')
-pm = activity.getPackageManager()
-flash_available = pm.hasSystemFeature(PackageManager.FEATURE_CAMERA_FLASH)
+from plyer.platforms.android import SDK_INT
+if SDK_INT >= 23:
+  Context = autoclass('android.content.Context')
+  CameraManager = activity.getSystemService(Context.CAMERA_SERVICE)
+  CameraCharacteristics = autoclass("android.hardware.camera2.CameraCharacteristics")
+  CamId = CameraManager.getCameraIdList()[0] # First Camera
```
kuzeyron, line 15: Personally I don't like this method to force it to be [0]. For loop with checks would be one way of dealing with this. All Androids do not have a camera so it would be basically better to find out if it's supported.

PR #807, opened 2024-04-27
replace old mock by unittest.mock

a-detiste:
mock is now part of the Python standard library, available as [unittest.mock](https://docs.python.org/dev/library/unittest.mock.html) in Python 3.3 onwards.

https://github.com/testing-cabal/mock


misl6 (approved):
LGTM. Thank you!

PR #811, opened 2024-05-23
Title parameter should be 'title' instead of 'name' (zenity)

Monsterovich:
**Correct:**
`zenity --file-selection --title='My title'`
**Incorrect:**
`zenity --file-selection --text='My title'`


akshayaurora:
Fixed in #828

PR #822, opened 2024-10-23
remove unsupported options for Zenity

### LINKED ISSUE #821: Linux Zenity on Ubuntu 24 no longer supports --name property (and other properties)
Zenity file chooser dropped the --name option, which causes the following error when using the file chooser:

```
/usr/bin/zenity --file-selection --filename /home/brent/tmp --name "Load Configuration" --file-filter *.rcp
This option is not available. Please see --help for all possible usages.
```

Suggestion: drop the --name option under all circumstances to provide forward and backwards compatibility

brentpicasso:
Resolves #821 

* --file-selection is now obsolete
* --name option in zenity file chooser was abruptly dropped starting in Ubuntu 24 (which prompted this PR)
* --window-icon option was dropped (unknown when, but would also cause it to fail)

```
/usr/bin/zenity --help-file-selection
Usage:
  zenity [OPTION…]

File selection options
  --file-selection                                  Display file selection dialog
  --filename=FILENAME                               Set the filename
  --multiple                                        Allow multiple files to be selected
  --directory                                       Activate directory-only selection
  --save                                            Activate save mode
  --separator=SEPARATOR                             Set output separator character
  --file-filter=NAME | PATTERN1 PATTERN2 ...        Set a filename filter
  --confirm-overwrite                               DEPRECATED; does nothin
```

### DIFF
--- a/plyer/platforms/linux/filechooser.py
+++ b/plyer/platforms/linux/filechooser.py
@@ -106,8 +106,7 @@ class ZenityFileChooser(SubprocessFileChooser):
     def _gen_cmdline(self):
         cmdline = [
             which(self.executable),
-            "--file-selection",
-            "--confirm-overwrite"
+            "--file-selection"
         ]
         if self.multiple:
             cmdline += ["--multiple"]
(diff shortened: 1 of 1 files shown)


misl6 (approved):
LGTM. Thank you!

brentpicasso:
Happy to help! <3

PR #823, opened 2024-11-30
feat: add VOIP with permission verification to Kivy

SanquezH:
- Integrate VOIP functionality for Android and iOS developers
- Supports Android to iPhone two-way communication
- Implements permission verification for microphone access
- Allows for auto or manual selection of TLS protocols to secure VOIP call
- Gracefully ends call automatically if the connection drops unexpectedly
- Incorporates Swift and Objective-C code in plyer through custom frameworks

Android & iOS Demo Video: https://youtu.be/9GQdY-4Iuz4
Original Project: https://github.com/SanquezH/Kivy-VOIP


SanquezH:
This is my first time submitting a change, and I just wanted to check in on the status of this PR. If there’s anything I can do to help or adjust the changes, please let me know.

kuzeyron:
Looks like you are missing out to fill in for other OS's so it will throw a "not implemented" error. If you are able to implement this code with other OS's it would be appreciated if it did!

SanquezH:
I specialize in Android app development, which is why this code is currently designed for Android only. While other OSs aren’t supported at the moment, I’m open to future collaboration to implement iOS support. I’ve recently started exploring iOS development with Kivy after getting a MacBook two weeks ago, so I’m still catching up.

kuzeyron:
> Looks like you are missing out to fill in for other OS's so it will throw a "not implemented" error.

Still remains. Have a look on how other implementations are doing. Take your time :)

SanquezH:
The temperature, humidity, and IR Blaster modules are currently supported only on Android. What steps can I take to ensure the same level of acceptance?

kuzeyron:
In `plyer/platforms/android/voip.py` you should remove the platform check because Plyer is already doing that for you.

SanquezH:
@kuzeyron The platform check is removed in the latest commit change.

ElliotGarbus:
@SanquezH My hope is that these comments are helpful to you in getting this PR accepted.  I'm not a maintainer, I'm also not an expert at plyer - but I see a few things that need attention.

The code does not seem to "match" the plyer framework.  Lets use accelerometer as an example.  The facade defines a parent class that provides a platform independent interface to an accelerator.   The private and public interfaces are defined.  You have added a new interface, enable_debug() this is unique to VOIP, and does not match the rest of the framework.  I'd recommend removing this.   

You also have a number of class variables in the facade.  Looking across the library, I do not see this done in the facade, instead the facade defines a property (@property), and uses calls to access these kinds of attributes.  See the implementation of acceleration in facades/accelerometer.py.  Given I don't see this kind of use of class vars in any of the existing capabilities, it would seem most if not all of these attributes could be parameters to a make_call() method, matching the call implementation.  

The implementation of VOIP does not work with the facade.  Again lets look at the call class as an an example.  The android implementation of the Call class is names AndroidCall, and derived from the facades Call class.  The VOIP implementation needs to match this structure so the facades can dynamically select the appropriate implementation for the target platform.  Also note in android/call.py there is a function instance() that returns an instance of Android call.  This is consistent in all of the platform files, and required for the proper implementation of the library.  

To better understand the Plyer library take a look at plyer/__init__.py  You can see that the Proxy class is instanced for each "component" in the library.   Looking at plyer/plyer/utils.py we can see the Proxy class implementation.   This code is rather sophisticated, looking at the _ensure_obj() method we can see that instance() function is called here from the appropriate file.  The __getattribute__ method will call ensure_obj() when an attribute the create proxy class is made.

Hope that helps - thanks for making this contribution!

ElliotGarbus:
> The temperature, humidity, and IR Blaster modules are currently supported only on Android. What steps can I take to ensure the same level of acceptance?

Look at the facade for irblaster, you will see the private methods raise a NotImplemented() Error.  You will want to add this to the VOIP facade.  When the AndroidVOIP is instanced, these will be overridden.  For platforms that do not have an implementation - the exception will be raised.

SanquezH:
@ElliotGarbus Thank you for your fee

SanquezH:
@ElliotGarbus & @kuzeyron I think the VOIP module is finally incorporated into the Plyer module as desired. What do you think?

Facade: https://github.com/SanquezH/plyer/blob/VOIP/plyer/facades/voip.py
Android Platform Code: https://github.com/SanquezH/plyer/blob/VOIP/plyer/platforms/android/voip.py
plyer __init__.py: https://github.com/SanquezH/plyer/blob/VOIP/plyer/__init__.py

ElliotGarbus:
@SanquezH Overall very well done.  

Here are a few things to consider that I expect the maintainer will have a strong opinion on.  I don't know the direction for the library - so consider these food for thought.  If the direction is not to have logging in the lib (as it appears now - you could make the self.debug a private option to get the logging messages. 

All of your log messages are debug.  I wonder the messages that get turned on if the debug parameter is set should be info level messages.  They seem more informational, and would not require the logging level to be set to debug to be captured.  There are places where you are capturing exceptions, it would make sense to me to have these logged as exceptions - independent of the debug flag.

SanquezH:
@ElliotGarbus Thank you for helping me throughout the process, I learned a lot from this. I will change the log levels to info or error where applicable. In the change I submitted about 45 minutes ago, self.debug is a private variable. If the maintainers prefer the optional logs to be removed, I will gladly comply. However, for now, I would like to retain the logs to provide developers with an easy way to identify issues directly, without relying on external sources.

ElliotGarbus:
@SanquezH You're welcome.  Your choices make sense to me.  Note there is a log message for exceptions.   Good Luck!

SanquezH:
The latest updates to AndroidVoip and the VOIP example improve error handling, allowing developers to identify critical issues and present them to users through dialogs or messages in the following scenarios:
• Microphone permission errors: Check the voip.hasPermission boolean variable in the example.
• Unreachable server errors: Check the voip.connected boolean variable in the example.

I have tested the code, and it performed as expected. I will refrain from making any further changes unless directed by a maintainer.

SanquezH:
@kuzeyron Is it possible to also add a testing label for platform-ios?

kuzeyron:
Awesome! It's now up to the core devs if it's now okay.

SanquezH:
I created a video demonstrating how this module transmits audio data with visual representation. https://youtu.be/9GQdY-4Iuz4

PR #824, opened 2024-11-30
Set "DefExt" argument in Win32FileChooser run method.

ByteArts:
Set value for the "DefExt" key so that Windows will append the extension to the returned file name.


akshayaurora (approved):

akshayaurora:
@ByteArts thanks

PR #832, opened 2025-02-25
 feat: add VoIP with permission verification

SanquezH:
- Integrate VOIP functionality for Android and iOS developers
- Supports Android to iPhone two-way communication
- Implements permission verification for microphone access
- Allows for auto or manual selection of TLS protocols to secure VOIP call
- Gracefully ends call automatically if the connection drops unexpectedly
- Incorporates Swift and Objective-C code in plyer through custom frameworks

Android & iOS Demo Video: https://youtu.be/9GQdY-4Iuz4
Original Project: https://github.com/SanquezH/Kivy-VOIP


SanquezH:
@kuzeyron @akshayaurora @ElliotGarbus @misl6  I think that the Ubuntu test failure in continuous integration was due to the capitalization of the VoIP branch name. While Kivy-iOS tolerated the capitalization, Buildozer required it to be lowercase, which is why I renamed the branch.

I had previously tested the VoIP module on Android using a local copy of the git repository, which is why it worked locally. However, it failed when installing via pip git in buildozer.spec requirements due to the branch name capitalization. Lowercasing the branch resolved the issue.

That said, no files were changed in this new pull request, only the branch name has been updated.

SanquezH:
@akshayaurora I now realize the issue is related to the coding style. I'm making the necessary corrections and will notify you once it's done for another review.

SanquezH:
@akshayaurora I've made the necessary changes to align with the Flake8 coding style, which should correct the style error.

akshayaurora (approved):

SanquezH:
@akshayaurora I successfully replicated the testing environment and reproduced the same error observed in the Ubuntu environment. The issue was caused by whitespace inconsistencies in multiline strings used for argument comparisons within the self.assertEqual function. While the character count was cutting it close, I reformatted the code to remove multiline strings while ensuring test_utils.py remains compliant with Flake8 style guidelines. I will continue applying this troubleshooting approach until all issues are fully resolved.

akshayaurora:
@SanquezH thanks for all your work, I am hesitating a bit cause of the inclusion of the framework of voip itself.

Would it be possible to not include the framework inside plyer ?

I absolutely understand that this framework is pretty integral to the feature.
My only concern is if many such framework get included by default with plyer while many people don't use such frameworks it could quickly lead to ballooning of size. 

Here in this particular instance I guess code is not that much.

Can't think of a better alternative though... so maybe it's fine to include the framework till we come up with a better alternative.

@AndreMiras @misl6 @Cheaterman thoughts ?

SanquezH:
@akshayaurora Due to the limitation of including Objective-C blocks in Pyobjus, I created the framework as a workaround for the installtaponbus function. Since the framework is essential for VoIP functionality on iOS, the code would not function as designed without it. I could potentially relocate it to another library, such as a frameworks directory within the Pyobjus repository. However, that would require waiting for its acceptance there before referencing it in this pull request. Let me know if there's anything I can provide to address your concerns.

akshayaurora:
@SanquezH no need, I think it should be fine to include for now, if it becomes a issue later on, we could take a look again, when we cross that road.

SanquezH:
@akshayaurora I will think of a solution in the future. At the moment, I'm waiting for the virtual machine I set up on my MacBook to load. Once it's ready, I'll create a container to replicate the testing environment for the OSX failure.

SanquezH:
@akshayaurora I removed the use of indentations and newlines in the comparison between the decorator in utils.py and the string comparison in test_utils.py during the self.assertEqual function call, allowing the tests to run successfully locally on my Ubuntu and OSX environment.

akshayaurora:
@SanquezH thanks a lot  💯

akshayaurora:
@SanquezH How about some documentation for people who would like to remove the said framework ? Would be helpful.

SanquezH:
@akshayaurora, you're very welcome! I truly appreciate your team for giving me the opportunity to make my first contribution to the software development world. I'll work on finding a convenient way to remove the frameworks if desired. Where would you like me to store the documentation?

akshayaurora:
Documenting in the plyer docs in the Readme.md should be good 👍

SanquezH:
@akshayaurora Great, I'll include it in a new pull request once I find the solution.

SomberNight:
It would have been nice to squash this branch to ~5 commits, down from 66, before merging. It is full of 1-2 line style fixes and trivial follow-ups.

Also, what's up with the included binary files, e.g. `plyer/platforms/ios/frameworks/simulator/Voip.framework/Voip` is a 150 KB binary (and there are 8 other binary files).
How were those files made?

akshayaurora:
@SomberNight I assume your concern is about inclusion in a software that cares about security.

There is a growing need for plyer to include binaries / remote frameworks like this one.

Probably a better idea here would have been to include documentation about how to compile said framework from source? or to include the sources/compile instruction as a sub-module under git ?

If that is your concern maybe we can do that to better define for future on how to include binaries. (Basically I am offering this as a solution,  and that I would look into replacing this framework in one of the manners mentioned above so source is easier to authenticate... )

SanquezH:
I’ll make sure future branches are squashed so the commit history stays concise and easy to review. The binary files were included to support the iOS implementation and were built using Xcode on macOS. I plan to gather the steps taken to build these binaries and include the source code files along with a BUILD.md soon, so anyone can reproduce the framework and generate their own compiled binaries. For convenience, I plan to keep the prebuilt binaries available, while also providing the source and build instructions to maintain transparency and reproducibility. Does this plan work for you, @akshayaurora ?

PR #836, opened 2025-03-24
Some plyer improvements

GvozdevLeonid:
- Added the ability to get GPS coordinates for macos
- Fixed some wifi errors on macos
- Added the ability to get GPS coordinates on macos
- Changed the function of getting the device name for android and macos. My branch also contains code for getting the device name on ios, but I could not check it because I could not build the kivy-ios application


plyer/platforms/android/orientation.py
```diff
@@ -38,6 +39,21 @@ def _set_sensor(self, **kwargs):
             activity.setRequestedOrientation(
                 ActivityInfo.SCREEN_ORIENTATION_SENSOR_PORTRAIT)
 
+    def _get_orientation(self):
+        rotation = activity.getWindowManager().getDefaultDisplay().getRotation()
+
+        if rotation == Surface.ROTATION_0:
```
kuzeyron: How about something like this?:
```
surfaces = {Surface.ROTATION_90: 'landscape', ...}
return surfaces.get(rotation, 'unknown')
```

plyer/platforms/linux/orientation.py
```diff
@@ -14,14 +14,41 @@ def _set_landscape(self, **kwargs):
...
+
+        try:
+            orientation = sb.check_output(
+                f"xrandr -q --verbose | grep {self.screen} | sed 's/primary //' | awk '{{print $5}}'",
+                shell=True
+            ).decode('utf-8').strip()
+        except Exception:
+            return 'unknown'
+
+        if orientation == 'normal':
```
kuzeyron: How about something like this?
```
orientations = {'inverted': 'landscape-reversed', ...}
return orientations.get(orientation, 'landscape')
```

plyer/platforms/android/orientation.py
GvozdevLeonid: yes, that would be better

kuzeyron: Can you update this PR?

kuzeyron (approved):

kuzeyron:
LGTM! Thank you!

PR #838, opened 2025-04-11
Update linux/battery module to support both BAT0 and BAT1

### LINKED ISSUE #743: [Bug] battery.status gives an error.
Is in title, for any reason when I write `battery.status` the following error appears:

```python
  File "/usr/lib/python3.10/site-packages/plyer/utils.py", line 117, in __getattribute__
    result = getattr(object.__getattribute__(self, '_obj'), name)
  File "/usr/lib/python3.10/site-packages/plyer/facades/battery.py", line 42, in status
    return self.get_state()
  File "/usr/lib/python3.10/site-packages/plyer/facades/battery.py", line 49, in get_state
    return self._get_state()
  File "/usr/lib/python3.10/site-packages/plyer/platforms/linux/battery.py", line 81, in _get_state
    percentage = float(
ValueError: could not convert string to float: '0% (should be ignored'

```

Edit: I'm using Arch Linux

zencraftr:
# Update linux/battery module to support both BAT0 and BAT1

## Description

This PR added support for **BAT1** in the linux battery module. For some linux dirstros, such as Arch Linux (which I run) the battery for certain laptop can be **BAT1**, instead of **BAT0** which `plyer` already supported.

This PR will make `plyer` battery module work for every linux distribution. Fixing the error I asked to reopen and now Closes #743.

## Related Issue(s)

- Related to #743 
There might be other issues closed, like this past one but not resolved.

## Changes

- Updated linux/battery module to support **BAT1**.
- Refactored conflicting code in `plyer/tests/test_battery.py`, which was running `assert_called_once_with()` and added testing for both **BAT0** and **BAT1**. N.B.: I didn't understand why had to make sure it was called once, please refactor or lemme know.

## How Has This Been Tested?

- Ran all unit tests using `pytest` and confirmed they passed, but `TestUniqueID.test_uniqueid` which was failing beforehand.
- Performed manual testing on Linux to verify  BAT1 work as expected. **Need somebody to manually test on a Linux system with BAT0**

## Screenshots

plyer-bat1

## Checklist

- [x] My code follows the project's style guidelines, I hooked PEP8 before committing.
- [x] I have commented my code, especially in hard-to-understand areas.
- [x] Change documentation is not needed as it doesn't specify the if the battery is **BAT0** or **BAT1**
- [x] My changes generate no new warnings.
- [x] I have added tests to prove my fix or feature works.
- [x] Any dependent changes have been merged and published in downstream modules.

## Additional Notes

The only concern is that I had to change the previous test as it required for the battery to only be called once (I don't understand why), the command is one call already.
However, if there is a reason behind it, this my port breaking changes. Please @KeyWeeUsr, let me know if the check for one call is mandatory.

### DIFF
--- a/plyer/platforms/linux/battery.py
+++ b/plyer/platforms/linux/battery.py
@@ -19,7 +19,21 @@ class LinuxBattery(Battery):
     def _get_state(self):
         status = {"isCharging": None, "percentage": None}
 
-        kernel_bat_path = join('/sys', 'class', 'power_supply', 'BAT0')
+        # Try BAT0 first, then BAT1
+        bat_names = ['BAT0', 'BAT1']
+        kernel_bat_path = None
+
+        for bat_name in bat_names:
+            path = join('/sys', 'class', 'power_supply', bat_name)
+            if exists(path):
+                kernel_bat_path = path
+                self.battery_name = bat_name
+                break
+        
+        # If no battery found, return empty status, might need to be implemented
+        if not kernel_bat_path:
+            return status
+
         uevent = join(kernel_bat_path, 'uevent')
 
         with open(uevent) as fle:
@@ -54,16 +68,22 @@ class UPowerBattery(Battery):
         environ['LANG'] = 'C'
         status = {"isCharging": None, "percentage": None}
 
-        # We are supporting only one battery now
-        # this will fail if there is no object with such path,
-        # however it's safer than 'upower -d' which provides
-        # multiple unrelated 'state' and 'percentage' keywords
-        dev = "/org/freedesktop/UPower/devices/battery_BAT0"
-        upower_process = Popen(
-            ["upower", "--show-info", dev],
-            stdout=PIPE
-        )
-        output = upower_process.communicate()[0].decode()
+        # Try BAT0 first, then BAT1
+        bat_names = ['BAT0', 'BAT1']
+        output = None
+
+        for bat_name in bat_names:
+            dev = f"/org/freedesktop/UPower/devices/battery_{bat_name}"
+            upower_process = Popen(
+                ["upower", "--show-info", dev],
+                stdout=PIPE
+            )
+            result = upower_process.communicate()[0].decode()
+            if result and "should be ignored" not in result.lower():
+                output = result
+                self.battery_name = bat_name
+                break
+            
         environ['LANG'] = old_lang
         if not output:
             return status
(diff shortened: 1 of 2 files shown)


plyer/platforms/linux/battery.py
```diff
@@ -19,7 +19,21 @@ class LinuxBattery(Battery):
     def _get_state(self):
         status = {"isCharging": None, "percentage": None}
 
-        kernel_bat_path = join('/sys', 'class', 'power_supply', 'BAT0')
+        # Try BAT0 first, then BAT1
+        bat_names = ['BAT0', 'BAT1']
```
AndreMiras, line 23: Great that makes sense in general, it's a better iteration over what we had previously.
But why not taking it a bit further?
What do you think of something like:
```python
def discover_batteries():
    """Dynamically discover all available batteries."""
    power_supply_path = '/sys/class/power_supply'
    batteries = []
    
    if not exists(power_supply_path):
        return batteries
    
    for item in os.listdir(power_supply_path):
        item_path = join(power_supply_path, item)
        type_file = join(item_path, 'type')
        
        if exists(type_file):
            with open(type_file, 'r') as f:
                if f.read().strip() == 'Battery':
                    batteries.append(item)
    
    return sorted(batteries)
```

plyer/platforms/linux/battery.py
```diff
@@ -19,7 +19,21 @@ class LinuxBattery(Battery):
...
+        kernel_bat_path = None
+
+        for bat_name in bat_names:
+            path = join('/sys', 'class', 'power_supply', bat_name)
+            if exists(path):
+                kernel_bat_path = path
+                self.battery_name = bat_name
+                break
```
AndreMiras, line 31: But in the end we only keep the first valid battery, meaning we don't really support multiple.
I guess it's still a better iteration over what we had, but in the long run the interface probably needs to evolve to support multiple

plyer/platforms/linux/battery.py
```diff
@@ -54,16 +68,22 @@ def _get_state(self):
...
-        # however it's safer than 'upower -d' which provides
-        # multiple unrelated 'state' and 'percentage' keywords
-        dev = "/org/freedesktop/UPower/devices/battery_BAT0"
-        upower_process = Popen(
-            ["upower", "--show-info", dev],
-            stdout=PIPE
-        )
-        output = upower_process.communicate()[0].decode()
+        # Try BAT0 first, then BAT1
+        bat_names = ['BAT0', 'BAT1']
```
AndreMiras, line 72: Same as https://github.com/kivy/plyer/pull/838/files#r2342198194

AndreMiras:
Thanks for the change, I don't usually keep an eye on this repo and I missed it.
Nice addition overall :heart:  I made a few suggestions, let us know what you think

PR #854, opened 2026-09-13
fix battery on macOS Silicon

rshah713:
Currently fails with:
```python
>>> battery.status
  File "<python-input-1>", line 1, in <module>
    battery.status
    ...truncated error...
    max_capacity = float(line.rpartition('=')[-1].strip())
ValueError: could not convert string to float: '29}'
```

This is because on Silicon MacBooks, `ioreg -rc AppleSmartBattery` has nested `IsCharging`, `MaxCapacity`, `CurrentCapacity`, etc under `BatteryData`:

## Raw Output
```
developer@MacBook-Pro ~ % ioreg -rc AppleSmartBattery | grep MaxCapacity
 "BatteryData" = {"Ra03"=34,"Ra10"=42,"CellWom"=(0,0),"Ra14"=274,"MaxCapacity"=100,"ChemicalWeightedRa"=0,...}

    "MaxCapacity" = 100
    "AppleRawMaxCapacity" = 8739

```

## Solution

Skip the lines of output until we obtain the last MaxCapacity which is simply outputted (similar to Intel-based implementation).

```python
>>> from plyer import battery
>>> battery.status
{'isCharging': False, 'percentage': 35.59496567505721}
```


plyer/platforms/macosx/battery.py
```diff
@@ -35,7 +35,10 @@ def _get_state(self):
             if 'IsCharging' in line:
                 is_charging = line.rpartition('=')[-1].strip()
             if 'MaxCapacity' in line:
```
kuzeyron, line 37: Since you added try/except. This if statement is completely useless.

PR #855, opened 2026-09-13
Bluetooth macOS fix

rshah713:
`bluetooth.info` returns None on macOS as `system_profiler SPBluetoothDataType` no output has changed from `Bluetooth Power: <on/off>` to `State: <on/off>`


plyer/platforms/macosx/bluetooth.py
```diff
@@ -28,7 +28,11 @@ def _get_info(self):
 
         lines = []
         for line in output:
-            if 'Bluetooth Power' not in line:
+            parts = line.split()
+            # Bluetooth Power: On/Off (older) or State: On/Off (newer).
+            if 'Bluetooth Power' not in line and not (
+                    'State:' in line and parts and
+                    parts[-1] in ('On', 'Off')):
```
kuzeyron, line 35: Instead of this it would be better if you just check if the three is in line. You can add a helper function to this section.

